PR #3, opened 2019-06-09
Add option to anonymize line numbers

### LINKED ISSUE #2: Provide a way to 'anonymize' line numbers when building a Snippet
In the rust test suite there's the option to replace line numbers in the diagnostics output with letters, so that changes in the output produce less noisy diffs.

With that flag enabled, the following

```
error: cannot concatenate a byte string literal
  --> $DIR/concat.rs:2:13
   |
 2 |     concat!(b'f');
   |             ^^^^
```

is turned into

```
error: cannot concatenate a byte string literal
  --> $DIR/concat.rs:2:13
   |
LL |     concat!(b'f');
   |             ^^^^
```

Any line number is replaced by `LL`, no matter the digits of that number.

AFAICT, rustc would currently have to implement its own `Snippet -> DisplayList` conversion, which would duplicate a lot of the conversion code just to anonymize the line numbers.

I think this would be nice to have inside `annotate-snippets` too, maybe as an option when constructing a Snippet? I believe other users of this library could also benefit from this option.

phansch:
With the current diagnostics emitter of Rust, the `-Z ui-testing` flag
allows to 'anonymize' line numbers in the UI test output.

This means that a line such as:

     2 |     concat!(b'f');

is turned into:

    LL |     concat!(b'f');

This is done because it makes the diff of UI test output changes much
less noisy.

To support this with `annotate-snippet`, we add a second parameter to
`DisplayListFormatter` that, when true, replaces the line numbers in the
left column with the text `LL`. The replacement text is always `LL` and
it does not affect the initial location line number.

In the new `annotate-snippet` emitter in rustc, we can then use this
parameter depending on whether the `-Z ui-testing` flag is set or not.

Closes #2 
cc https://github.com/rust-lang/rust/issues/59346

### DIFF
--- a/src/formatter/mod.rs
+++ b/src/formatter/mod.rs
@@ -20,9 +20,11 @@ fn repeat_char(c: char, n: usize) -> String {
     s.repeat(n)
 }
 
-/// DisplayListFormatter' constructor accepts a single argument which
-/// allows the formatter to optionally apply colors and emphasis
-/// using `ansi_term` crate.
+/// DisplayListFormatter' constructor accepts two arguments:
+///
+/// * `color` allows the formatter to optionally apply colors and emphasis
+/// using the `ansi_term` crate.
+/// * `anonymized_line_numbers` will replace line numbers in the left column with the text `LL`.
 ///
 /// Example:
 ///
@@ -30,7 +32,7 @@ fn repeat_char(c: char, n: usize) -> String {
 /// use annotate_snippets::formatter::DisplayListFormatter;
 /// use annotate_snippets::display_list::{DisplayList, DisplayLine, DisplaySourceLine};
 ///
-/// let dlf = DisplayListFormatter::new(false); // Don't use colors
+/// let dlf = DisplayListFormatter::new(false, false); // Don't use colors, Don't anonymize line numbers
 ///
 /// let dl = DisplayList {
 ///     body: vec![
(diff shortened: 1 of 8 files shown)


phansch:
One of my small gripes with Rust is that it doesn't have keyword arguments (yet?). That would make passing these boolean options much more understandable at the call sites. I'm not sure if there's currently a clearer way?

oli-obk:
> keyword arguments 

you can always create structs with appropriate field names ;)

src/formatter/mod.rs
```diff
@@ -75,7 +87,13 @@ impl DisplayListFormatter {
             DisplayLine::Source {
                 lineno: Some(lineno),
                 ..
-            } => cmp::max(lineno.to_string().len(), max),
+            } => {
+                if self.anonymized_line_numbers {
+                    2 // "LL"
```
zbraniecki: Should you take a len of `ANONYMIZED_LINE_NUM` instead?

phansch:
Replaced the magic constant.

@zbraniecki what are your feeling regarding the change of the api of the `DisplayListFormatter` constructor? I feel like it's still fine when passing two booleans. If there turns out to be a need for more configuration, it could be changed later?

zbraniecki:
> If there turns out to be a need for more configuration, it could be changed later?

Yep, it's an internal API, I'm fine with adjusting it later.

oli-obk:
I wonder...

@bors r+

oli-obk:
I guess not

phansch:
Thanks! I think this will also need a new release before it can be used in rustc, right?

phansch:
@zbraniecki (or someone else?) could you push a new version to crates.io?

zbraniecki:
yeah, will do.

zbraniecki:
done. 0.6.0 released.

zbraniecki:
@phansch did you intend to turn also context lines (without numbers) to `LL`?

I see this:

```
error[E0003]: Expected token: "="
 --> ../fluent-bundle/examples/resources/en-US/simple.ftl:2:18
  |
2 | input-parse-error Error: Could not parse input `{ $input }`. Reason: { $reason }
  |                   ^^ Expected token: "="
  |
-----------------------------
```

turned into:

```
error[E0003]: Expected token: "="
  --> ../fluent-bundle/examples/resources/en-US/simple.ftl:2:18
LL |
LL | input-parse-error Error: Could not parse input `{ $input }`. Reason: { $reason }
LL |                   ^^ Expected token: "="
LL |
-----------------------------
```

I think I'd expect:

```
error[E0003]: Expected token: "="
  --> ../fluent-bundle/examples/resources/en-US/simple.ftl:2:18
   |
LL | input-parse-error Error: Could not parse input `{ $input }`. Reason: { $reason }
   |                   ^^ Expected token: "="
   |
-----------------------------
```

instead.

phansch:
Yeah I noticed that, too. That's a bug of the current implementation. My current plan is to only show `LL` for `DisplaySourceLine::Content` and where `DisplayLine::Source` has a `lineno` set. I think that should be enough?

(I've been a bit time restricted the past month or so but I want to get back into it again this month)

PR #5, opened 2019-07-15
Fix too many anonymized line numbers

phansch:
This expands the logic for replacing line numbers with `LL` so that line
numbers are only replaced, when they are actually provided.

When `lineno` is `None`, no `LL` should be shown in the output.

### DIFF
--- a/tests/formatter.rs
+++ b/tests/formatter.rs
@@ -515,13 +515,26 @@ fn test_anon_lines() {
                 range: (0, 19),
             },
         },
+        DisplayLine::Source {
+            lineno: None,
+            inline_marks: vec![],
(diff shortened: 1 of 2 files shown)


tests/formatter.rs
```diff
@@ -515,13 +515,26 @@ fn test_anon_lines() {
                 range: (0, 19),
             },
         },
+        DisplayLine::Source {
+            lineno: None,
+            inline_marks: vec![],
+            line: DisplaySourceLine::Empty,
```
phansch: Should the case of `DisplaySourceLine::Empty` + `Some(lineno)` also be handled in some way? I didn't see any test cases for it, so I left that out for now.

PR #15, opened 2019-10-29
Add InlineAnnotation

zbraniecki:
This is an experimental concept to annotate source inline on top of annotating it below.


zbraniecki:
This is an experiment in extending the API to support inline annotations.

I'm not sure if it's worth it, but wanted to give it a try and got it to work!

Screen Shot 2019-10-29 at 2 16 01 PM

Alternatively, I was thinking of extending `SourceAnnotation` to allow users to specify the display format - `under`, `inline` or both, but that made the API a bit less readable IMO.

zbraniecki:
All I had to add to the examples/format.rs is:

Screen Shot 2019-10-29 at 2 17 54 PM

zbraniecki:
@CAD97, @wycats, @brendanzab - thoughts?

I think it may be a path forward to allow for source coloring, if we enable custom "styles" and allow for such annotations. Not sure if it's the best way, but it's an option.

My primary goal is to visually aid analyzing the slice, not syntax coloring.

epage:
#97 will expand the API to allow different types of annotations.

If someone wants to move this forward from where we are at, I'd recommend creating an issue to discuss the idea.

PR #20, opened 2019-12-09
print from main example in readme

yoshuawuyts:
I tried running the example locally but it didn't produce any output. Had to spend a bit of time
figuring out how to make output happen; not too hard, but figured this would save the next person trying this out from going through the trouble. Thanks!


oli-obk:
:+1: I went through the same "huh?" situation.

PR #30, opened 2020-04-01
Fix to no copy when format with yansi_term

botika:
This PR eliminates the need to reallocate while formatting. 
For this I have forked the `ansi_term` library to refactor to `fmt::Display` allowing to introduce flow with a closure that takes the current formatter as argument. 
It will depend on `yansi-term` instead of` ansi_term`, which I will keep.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,14 +16,14 @@ coveralls = { repository = "rust-lang/annotate-snippets-rs", branch = "master", 
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-ansi_term = { version = "0.12", optional = true }
+yansi-term = { version = "0.1", optional = true }
 
 [dev-dependencies]
 glob = "0.3"
 serde_yaml = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 difference = "2.0"
-ansi_term = "0.12"
+yansi-term = "0.1"
 criterion = "0.3"
 
 [[bench]]
@@ -32,4 +32,4 @@ harness = false
 
 [features]
 default = []
-color = ["ansi_term"]
+color = ["yansi-term"]
--- a/src/formatter/mod.rs
+++ b/src/formatter/mod.rs
@@ -1,7 +1,6 @@
 use std::{
-    cell::Cell,
     cmp,
-    fmt::{self, Display, Formatter, Write},
+    fmt::{self, Display, Write},
 };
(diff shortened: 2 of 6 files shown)


zbraniecki:
@botika what's your take on `yansi-term` vs `termcolor`? @brendanzab says that `termcolor` may be a better long term bet.

I'm happy to land this PR, but I'm also curious on your opinion on that :)

botika:
The library is very simple, just enter a prefix and a suffix. So it does not need much maintenance nor can it innovate significantly. With the fork in the `ansi_term` I have left the minimum to fulfill its function as quickly as possible

The reason for the PR is to remove all the unnecessary `mallocs` in the formatter.

zbraniecki:
I understand. I'm just curious on your take on long-term replacing `ansi_term` or `yansi_term` with `termcolor`?

botika:
It is more optimal implement `fmt::Display` than `io::Write`. It forces you to use his buffer or stream output. I think the only thing this library should do is put a prefix and a suffix around the format flow, and let the user use whatever writer they want. So you couldn't use `fmt::Formatter` for output stream. And I am unaware of compatibility outside of unix systems

botika:
I've been looking at `termcolor` in more depth, just as I expected, it's forced to reallocate, [here allocate to stack](https://github.com/BurntSushi/termcolor/blob/master/src/lib.rs#L1372-L1373) and [here allocate to writer]( https://github.com/BurntSushi/termcolor/blob/master/src/lib.rs#L1398) (wherever the writer writes). 


So, the main problems of `termcolor` are unnecessary use of memory and lack of abstraction of the writer

botika:
I updated the [write bench](https://github.com/botika/yarte/blob/master/benches/src/all.rs), the results between `io::Write` and `fmt::Display` are:
```
Teams io::Write          time:   [423.37 ns 430.62 ns 436.27 ns]
Teams fmt::Write         time:   [291.71 ns 292.90 ns 294.35 ns]     
```
- [implementation `io::Write`](https://github.com/botika/yarte/blob/master/benches/src/all.rs#L210-L263)
- [implementation `fmt::Display`](https://github.com/botika/yarte/blob/master/benches/src/all.rs#L147-L184)

The results are clear. The fastest way to write in Rust is to implement `fmt::Display`.

zbraniecki:
Thank you!

Issue #29, opened 2020-03-31
SourceAnnotation underline displays offset from range


gipsond:
I was getting #24 with v0.6.1, but not anymore. Unfortunately I got a similar problem when I tried to update to v0.7.0.

Here's with v0.6.1:
```
error: invalid immediate, couldn't parse number (was: 30_2)
 --> .\add.asm:1:6
  |
...
5 | .ORIG x30_2
  |       ^^^^^ invalid immediate here
  |
```
and with v0.7.0, removing the `DisplayListFormatter` and adding default `FormatOptions` to the snippet:
```
error: invalid immediate, couldn't parse number (was: 30_2)
 --> .\add.asm:1:11
  |
...
5 |   .ORIG x30_2
  |  ___________^
6 | | AND R1, R1, R1
  | |____^ invalid immediate here
  |
```
`add.asm` has 4 lines before that error, each ending in CRLF. It seems like the offset is in the other direction now.

_Originally posted by @gipsond in https://github.com/rust-lang/annotate-snippets-rs/issues/24#issuecomment-606806438_

Nadrieril:
Oh, #27 didn't fix the bug properly. As I said [here](https://github.com/rust-lang/annotate-snippets-rs/issues/24#issuecomment-588154407), one cannot use `.lines()` for this.
The relevant change in #27 is that they removed the "+1" in this line: https://github.com/rust-lang/annotate-snippets-rs/blob/7e73a1bdaa75986dafb33454bd1207fd84e576e9/src/display_list/from_snippet.rs#L186
What this does is that instead of treating each line ending as 2 chars, they treat it as 1 char. But the correct fix is to distinguish between the different types of line endings.
Generally I don't think char indexing is really what you want anyways, because unicode makes everything complicated.

botika:
@Nadrieril You can post the `Snippet` and expected output?

Nadrieril:
This is the exact same issue as in https://github.com/rust-lang/annotate-snippets-rs/issues/24, except that in #24 it failed on `\n` line endings and succeeded on `\r\n` endings, but now it's the reverse. I don't have code that triggers this available, but any code with a line ending in `\r\n` and a snippet pointing somewhere after that line should show the error.

Nadrieril:
Actually, here you go:
```rust
Snippet {
    title: Some(
        Annotation {
            id: None,
            label: Some(
                "oops".to_string(),
            ),
            annotation_type: AnnotationType::Error,
[...]

PR #32, opened 2020-04-03
Fix no copy when cast `Snippet` to `DisplayList` and #29

botika:
With this pull request the "no copy" process is completed. All the crate is free copy :smile: 
I have changed from `yaml` to `toml` because `yaml` does not implement `serde::Deserialize`


botika:
Fixed #29

src/display_list/from_snippet.rs
```diff
@@ -2,6 +2,52 @@
...
+        CursorLines(src)
+    }
+}
+
+#[repr(u8)]
+enum EndLine {
+    EoF = 0,
+    One = 1,
+    Two = 2,
+}
```
zbraniecki, line 18: Why do you use `repr(u8)` on this?
Also, can you name it CRLF and LF respectively? EOF is also all capitals.

src/display_list/from_snippet.rs
```diff
@@ -236,9 +282,7 @@ fn format_body(slice: snippet::Slice<'_>, has_footer: bool) -> Vec<DisplayLine<'
     let mut current_index = 0;
     let mut line_index_ranges = vec![];
 
-    let lines = slice.source.lines();
-    let lines_len = lines.clone().count();
-    for (i, line) in lines.enumerate() {
+    for (line, end_line) in CursorLines::new(slice.source) {
```
zbraniecki, line 285: I really like this solution!

src/display_list/from_snippet.rs
botika: It's done

zbraniecki:
Looks great! Thank you!

PR #34, opened 2020-05-12
Removing duplicate Slice origins and surrounding empty lines

### LINKED ISSUE #9: Removing duplicate Slice origins and surrounding empty lines
I'm currently trying to move the suggestions of the following diagnostic over from Rust to annotate-snippet:

```
error[E0382]: use of moved value: `x`
  --> /code/rust/src/test/ui/annotate-snippet/suggestion.rs:4:5
   |
 4 |     let x = vec![1];
   |         - move occurs because `x` has type `std::vec::Vec<i32>`, which does not implement the `Copy` trait
 7 |     let y = x;
   |             - value moved here
 9 |     x; //~ ERROR use of moved value
   |     ^ value used here after move
```

In short, from Rust I can only get a `Vec` that contains the three lines with the annotations. This `Vec` contains _only_ the lines with annotations and no other lines around it. Something similar to this:

```
[
  Line {
      line_index: 4,
      annotations: [
          Annotation {
              start_col: 8,
              end_col: 9,
              label: "move occurs because ...",
          },
      ],
  },
  Line {
      line_index: 7,
      annotations: [
          Annotation {
              start_col: 11,
              end_col: 12,
              label: "value moved here",
          },
      ],
  },
  // etc..
]
```

I'm currently mapping this to annotate-snippets by turning each `Line` into a `Slice` and then calling `DisplayList::from(snippet)` at the end.

The problem is that each `Slice` has the `origin` at the top:

```
error[E0382]: use of moved value: `x`
  --> /code/rust/src/test/ui/annotate-snippet/suggestion.rs:4:8
   |
 4 |     let x = vec![1];
   |         ^ move occurs because `x` has type `std::vec::Vec<i32>`, which does not implement the `Copy` trait
   |
  ::: /code/rust/src/test/ui/annotate-snippet/suggestion.rs:7:12
   |
 7 |     let y = x;
   |             ^ value moved here
   |
  ::: /code/rust/src/test/ui/annotate-snippet/suggestion.rs:9:4
   |
 9 |     x; //~ ERROR use of moved value
   |     ^ value used here after move
```

If I set the origin of consecutive slices to `None` there will still be extra empty padding lines:

```
error[E0382]: use of moved value: `x`
  --> /code/rust/src/test/ui/annotate-snippet/suggestion.rs:5:8                                                             
   |
LL |     let x = vec![1];
   |         ^ move occurs because `x` has type `std::vec::Vec<i32>`, which does not implement the `Copy` trait
   |
   |
LL |     let y = x;
   |             ^ value moved here
   |
   |
LL |     x; //~ ERROR use of moved value
   |     ^ value used here after move
   |
```

I'm not sure what the original idea for a `Slice` was, but it seems like the best approach for those rustc annotations?

I can see three ways to solve this currently:

1. annotate-snippets doesn't add empty padding lines if the `origin` is `None`
2. annotate-snippets removes the origin and empty padding lines of consecutive Slices that have the same origin. This means annotate-snippet would have to compare the `origin` of each Slice to origin of the previous Slice.
3. We implement `From<Snippet> for DisplayList` by ourselves in rustc.

I would love to hear your thoughts on this before I continue with the annotations =)

cc https://github.com/rust-lang/rust/issues/61809

botika:
Closes #9 
@phansch ??

### DIFF
--- a/src/display_list/from_snippet.rs
+++ b/src/display_list/from_snippet.rs
@@ -104,15 +104,15 @@ fn format_annotation(annotation: snippet::Annotation<'_>) -> Vec<DisplayLine<'_>
 }
 
 fn format_slice(
-    mut slice: snippet::Slice<'_>,
+    slice: snippet::Slice<'_>,
     is_first: bool,
     has_footer: bool,
 ) -> Vec<DisplayLine<'_>> {
     let main_range = slice.annotations.get(0).map(|x| x.range.0);
-    let row = slice.line_start;
-    let origin = slice.origin.take();
+    let origin = slice.origin;
+    let line_start = slice.line_start;
     let mut body = format_body(slice, has_footer);
-    let header = format_header(origin, main_range, row, &body, is_first);
+    let header = format_header(origin, main_range, line_start, &body, is_first);
     let mut result = vec![];
 
     if let Some(header) = header {
--- a/src/display_list/from_snippet.rs
+++ b/src/display_list/from_snippet.rs
@@ -111,7 +111,8 @@ fn format_slice(
     let main_range = slice.annotations.get(0).map(|x| x.range.0);
     let origin = slice.origin;
     let line_start = slice.line_start;
-    let mut body = format_body(slice, has_footer);
+    let need_empty_header = origin.is_some() || is_first;
+    let mut body = format_body(slice, need_empty_header, has_footer);
     let header = format_header(origin, main_range, line_start, &body, is_first);
     let mut result = vec![];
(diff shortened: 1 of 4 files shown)


botika:
It seems that the changes in Rust are finally going to merge. And I would like to continue with this to adapt this crate and integrate it into rust. Does not appear @phansch . I will continue without him.

 Can @zbraniecki  review it, please?

zbraniecki:
Yes, thank you! It looks great!

PR #37, opened 2020-08-24
Fix character split when strip code

botika:
I had made an error to strip code in margin. In cases such as coverage, it could panic.

### DIFF
--- a/src/formatter/mod.rs
+++ b/src/formatter/mod.rs
@@ -1,6 +1,7 @@
 use std::{
     cmp,
     fmt::{self, Display, Write},
+    iter::once,
 };
 
 pub mod style;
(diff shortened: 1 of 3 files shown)


src/formatter/mod.rs
```diff
@@ -217,16 +218,21 @@ impl<'a> DisplayList<'a> {
                     } else {
                         false
                     };
+                    let mut ended = false;
```
zbraniecki, line 221: can you document this variable? It's a bit unique.

I also feel like there must be a way to make it simpler. The `range.0` is always the result of `acc.0`, so basically the first `take()`, and the `range.1` should be chainable `Option` on last element of `take_while` so if you put the iterator as a mutable variable, you should be able to do `iter.next().or(last)`.

The refactor is optional, and I'm happy to land this as-is with a comment explaining the variable and its role.

botika: This `ended` specifies that it will end in the next character so it will return until the next one to the final condition. In order to measure this last character, when end condition is true, in an efficient way. 

I will try to document it.

zbraniecki: I think I understand the role, but I had to run an iterator in my head to derive to that conclusion so a doc string would help.

PR #46, opened 2021-09-04
Fix display of annotation for double width characters

magurotuna:
Hi there.
We're using this crate in [denoland/deno_lint](https://github.com/magurotuna/deno_lint) to display pretty diagnostics of the linter. By chance, I found that this crate doesn't work well if the source includes double width characters like:

```ts
const a = "こんにちは";
```

Specifically, if we attempt to point to `は` in this snippet by passing `(15, 16)` as range, annotate-snippets will give us:

```
  |
1 | const a = "こんにちは";
  |                ^
  |
```

where obviously `^` points to the wrong position.
These characters take up twice the size of ordinary characters like the alphabet, so displaying annotation for them requires special handling, which I implemented in this patch.

### DIFF
--- a/tests/formatter.rs
+++ b/tests/formatter.rs
@@ -550,3 +550,31 @@ fn test_i_29() {
 
     assert_eq!(DisplayList::from(snippets).to_string(), expected);
 }
+
+#[test]
+fn test_point_to_double_width_characters() {
+    let snippets = Snippet {
+        slices: vec![snippet::Slice {
+            source: "こんにちは、世界",
+            line_start: 1,
+            origin: Some("<current file>"),
+            annotations: vec![snippet::SourceAnnotation {
+                range: (6, 8),
+                label: "world",
+                annotation_type: snippet::AnnotationType::Error,
+            }],
+            fold: false,
+        }],
+        title: None,
+        footer: vec![],
+        opt: Default::default(),
+    };
+
+    let expected = r#" --> <current file>:1:7
+  |
+1 | こんにちは、世界
+  |             ^^^^ world
+  |"#;
+
+    assert_eq!(DisplayList::from(snippets).to_string(), expected);
+}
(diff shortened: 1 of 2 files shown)


tests/formatter.rs
```diff
@@ -550,3 +550,98 @@ fn test_i_29() {
...
+        opt: Default::default(),
+    };
+
+    let expected = r#" --> <current file>:1:1
+  |
+1 | お寿司
+  | ^^^^^^ Sushi1
+2 | 食べたい🍣
+  |     ---- note: Sushi2
+  |"#;
```
magurotuna, line 644: With monospace fonts, this is displayed like:

image

tests/formatter.rs
```diff
@@ -550,3 +550,98 @@ fn test_i_29() {
...
+        opt: Default::default(),
+    };
+
+    let expected = r#" --> <current file>:1:3
+  |
+1 |   おはよう
+  |  _____^
+2 | | ございます
+  | |______^ Good morning
+  |"#;
```
magurotuna, line 607: image

tests/formatter.rs
```diff
@@ -550,3 +550,98 @@ fn test_i_29() {
...
+        title: None,
+        footer: vec![],
+        opt: Default::default(),
+    };
+
+    let expected = r#" --> <current file>:1:7
+  |
+1 | こんにちは、世界
+  |             ^^^^ world
+  |"#;
```
magurotuna, line 577: image

tests/formatter.rs
```diff
@@ -550,3 +550,126 @@ fn test_i_29() {
...
+        title: None,
+        footer: vec![],
+        opt: Default::default(),
+    };
+
+    let expected = r#" --> <current file>:1:7
+  |
+1 | こんにちは、新しいWorld！
+  |             ^^^^^^^^^^^ New world
+  |"#;
```
magurotuna, line 672: image

zbraniecki:
Thank you so much for the patch! It reads very clear and not just fixes the bug, but also improves the code.

zbraniecki (approved):

PR #47, opened 2021-09-04
Switch Travis CI URL back to .org

atouchet:
The .com URL now returns "No Results Found". Not sure if there's a better solution here.


zbraniecki:
We just need to switch to github actions :) Filed #48

zbraniecki (approved):

PR #54, opened 2022-01-03
Remove parsing of __ in title strings, fixes #53

### LINKED ISSUE #53: __ gets lost from labels
The starlark-rust library uses this code, specifically at https://github.com/facebookexperimental/starlark-rust/blob/13bec61a44dd2ec268403e1d6da4401ea76a3f12/starlark/src/errors/mod.rs#L205-L219. The relevant fragment is:

```rust
    let snippet = Snippet {
        title: Some(Annotation {
            label: Some(annotation_label),
            id: None,
        }),
        footer: Vec::new(),
        slices: slice.map(|s| vec![s]).unwrap_or_default(),
        opt: FormatOptions {
            color,
            ..Default::default()
        },
    };

    DisplayList::from(snippet)
```

However, if we have a label containing `__`, then that text gets lost from the label. The reason is that `format_label` in https://github.com/rust-lang/annotate-snippets-rs/blob/542e41e9a767c1c294564d549c46b2c3974b6481/src/display_list/from_snippet.rs#L56 splits by `__` to apply italics to alternate parts of the label. That behaviour was super surprising. I can't find it documented. There seems to be no way to turn it off, and no way to apply escaping to stop it happening. Can you suggest how I might get it so that errors don't get changed as they are rendered?

ndmitchell:
In #53 I reported the issue that `__` in titles gets parsed specially. This feature isn't documented anywhere, and doing a bit of git archaeology, I get lost 4 years ago with phrases like "more tests" adding the code. I can't figure out why someone would want this feature, but assume there was a reason when it was added, however, I'm not sure it is still useful. There doesn't seem any way to opt out of this feature, and as a result, error messages containing `__` get dropped (which is super confusing).

In the absence of feedback as the right way to proceed, I've made a diff removing the feature entirely. The alternative would be to document the feature, explain why it's useful, and explain how to avoid it (an alternative API? an escaping scheme?).

### DIFF
--- a/src/display_list/from_snippet.rs
+++ b/src/display_list/from_snippet.rs
@@ -53,22 +53,11 @@ fn format_label(
 ) -> Vec<DisplayTextFragment<'_>> {
     let mut result = vec![];
     if let Some(label) = label {
-        for (idx, element) in label.split("__").enumerate() {
-            let element_style = match style {
-                Some(s) => s,
-                None => {
-                    if idx % 2 == 0 {
-                        DisplayTextStyle::Regular
-                    } else {
-                        DisplayTextStyle::Emphasis
-                    }
-                }
-            };
-            result.push(DisplayTextFragment {
-                content: element,
-                style: element_style,
-            });
-        }
+        let element_style = style.unwrap_or(DisplayTextStyle::Regular);
+        result.push(DisplayTextFragment {
+            content: label,
+            style: element_style,
+        });
     }
     result
 }
--- a/tests/dl_from_snippet.rs
+++ b/tests/dl_from_snippet.rs
@@ -262,20 +262,10 @@ fn test_format_label() {
             annotation: dl::Annotation {
                 annotation_type: dl::DisplayAnnotationType::Error,
                 id: None,
-                label: vec![
-                    dl::DisplayTextFragment {
-                        content: "This ",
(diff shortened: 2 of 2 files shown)


ndmitchell:
I had a user of Starlark get really confused by the missing `__` recently. Anything we can do to get this merged?

jyn514:
I don't think this library is currently maintained.

zbraniecki:
I consider it maintained at low velocity. I plan to look into this issue over the weekend.

jshort:
I'm affected by this too so I'd love to see this merged!

harupy:
@zbraniecki I'm also affected by this. Can we merge this PR?

zong-zhe:
Hi, I want to ask when will this PR be merged? I have encountered the same issue. We have developed a language compiler 
[KCL](https://github.com/kcl-lang/kcl) that uses this library, and when the user-defined variable `__b` cannot be found, it is displayed as `a` in the error message, which is confusing.

zertosh:
Please consider merging this

zbraniecki:
Apologies for radio silence and thank you for the PR.

This feature has been added to support text style annotation based on the source, but I'm not sure if it's useful anymore.

zertosh:
@zbraniecki, no worries. Much appreciated. Any chance you can get a release out? Thank you :)

PR #61, opened 2023-11-08
Cleanup and updates

Muscraft:
This PR:

- Removes unneeded badges from `Cargo.toml` and `README.md`
- Moves to Edition `2021`
- Sets MSRV to `1.70.0` to [match CI](https://github.com/rust-lang/annotate-snippets-rs/blob/01ad63a88e53fd3cebe426f2ce8365609a4efb55/.github/workflows/ci.yml#L60)
- Sorted the dependencies alphabetically
- Fully specified dependency versions
- Updated `criterion` to the newest version 

Note: Updating `edition` and `rust-version` is ["Possibly" a semver-breaking](https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust) change. It is recommended that this is treated as a minor change, and since this is a `pre-1.0.0` crate, we should bump the patch version when this is released, that is `0.9.3`.

Note: I could not update `toml` to the newest version as it stopped supporting borrowing while deserializing, which is needed for the fixture tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "annotate-snippets"
 version = "0.9.2"
 dependencies = [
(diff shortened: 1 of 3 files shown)


zbraniecki:
> Note: I could not update toml to the newest version as it stopped supporting borrowing while deserializing, which is needed for the fixture tests.

I hit this too when I was planning to update. From reading the docs they are considering reverting that limitation. But since it's used in tests, we could do allocations if needed. For now I kept us stuck on 0.5

Issue #13, opened 2019-10-12
RFC: Proposal for "modern" API


CAD97:
In order to address concerns in #11, #12, https://github.com/wycats/language-reporting/issues/6, and probably others.

I've been experimenting with merging the APIs of `codespan`/`language-reporting`/`annotate-snippets`, and the below API surface is what that I think makes the most sense.

> NOTE: the suggested API has changed multiple times from feedback, see conversation starting at [this comment](https://github.com/rust-lang/annotate-snippets-rs/issues/13#issuecomment-543321203) for the most recent API and discussion.

<details><summary>Original Proposal</summary>

An experimental implementation of the API based on #12 is at https://github.com/CAD97/retort/pull/1 (being pushed within 24 hours of posting, I've got one last bit to "port" but I've got to get to bed now but I wanted to get this posted first).

> EDIT: I've reconsidered this API, though the linked PR does implement most of it. I'm sketching a new slightly lower-level design from this one, and the diagnostic layout of this current API will probably be a wrapper library around `annotate-snippets`. (I get to use the `retort` name!)

<details><summary>API</summary>

```rust
use termcolor::WriteColor;

trait Span: fmt::Debug + Copy {
    type Origin: ?Sized + fmt::Debug + Eq;

    fn start(&self) -> usize;
    fn end(&self) -> usize;
    fn new(&self, start: usize, end: usize) -> Self;
    fn origin(&self) -> &Self::Origin;
}

trait SpanResolver<Sp> {
    fn first_line_of(&mut self, span: Sp) -> Option<SpannedLine<Sp>>;
    fn next_line_of(&mut self, span: Sp, line: SpannedLine<Sp>) -> Option<SpannedLine<Sp>>;
    fn write_span(&mut self, w: &mut dyn WriteColor, span: Sp) -> io::Result<()>;
    fn write_origin(&mut self, w: &mut dyn WriteColor, origin: Sp) -> io::Result<()>;
}

#[derive(Debug, Copy, Clone)]
pub struct SpannedLine<Sp> {
    line_num: usize,
    char_count: usize,
    span: Sp,
}

impl Span for (usize, usize) {
    type Origin = ();
}

impl<Sp: Span<Origin=()>> Span for (&'_ str, Sp) {
    type Origin = str;
}

impl<Sp: Span> SpanResolver<Sp> for &str
where Sp::Origin: fmt::Display;

mod diagnostic {
    #[derive(Debug, Clone)]
[...]

PR #67, opened 2023-12-02
Move to a `Renderer` when for displaying a `Snippet`

Muscraft:
This PR moves to using `Renderer` as the external type for handling displaying a `Snippet`.

 This has a few advantages:
- Keeps what is doing the actual formatting as an internal implementation detail
  - In the future, different types of renderers could be used and exposed as an option to "pick"
- Moves the formatting options to be part of the renderer, not an individual `Snippet`
- Makes a clearer distinction between what is a `Snippet` and what is going to be "rendering" it

This PR also:
- moves to `anstyle` for color and removes the `color` feature
  -  `anstyle` is lightweight and writes nothing if the color is `Style::new`
  -  `anstyle` is what `cargo` uses and it would be good if when we use this crate we only need one color library.
- moves many things to be internal details

Note: this is a step in the direction of the proposals for a more modern API (#12 and #13). I have been taking inspiration from both PRs to get a clear vision of what the future should look like.

---

To review this PR, look at it commit by commit


epage:
Note that the biggest reason to use anstyle is that it is designed primarily to exist in public APIs for interoperability. Use for styling is secondary and there exist adapters to styling APIs

epage:
Note that the biggest reason to use anstyle is that it is designed primarily to exist in public APIs for interoperability. Use for styling is secondary and there exist adapters to styling APIs.

src/renderer/margin.rs
```diff
@@ -0,0 +1,115 @@
...
+        // ```
+
+        let mut m = Margin {
+            whitespace_left: whitespace_left.saturating_sub(6),
+            span_left: span_left.saturating_sub(6),
+            span_right: span_right + 6,
+            computed_left: 0,
+            computed_right: 0,
+            column_width,
+            label_right: label_right + 6,
```
zbraniecki, line 47: `nit`: can you hoist the magic numbers as consts?

src/renderer/margin.rs
```diff
@@ -0,0 +1,115 @@
...
+            } else {
+                self.computed_right
+            };
+        right < line_len && self.computed_left + self.column_width < line_len
+    }
+
+    fn compute(&mut self, max_line_len: usize) {
+        // When there's a lot of whitespace (>20), we want to trim it as it is useless.
+        self.computed_left = if self.whitespace_left > 20 {
+            self.whitespace_left - 16 // We want some padding.
```
zbraniecki, line 72: `nit`: hoist them to consts.

src/renderer/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::fmt::Display;
+
+#[derive(Clone)]
+pub struct Renderer {
+    anonymized_line_numbers: bool,
+    margin: Option<Margin>,
+    stylesheet: StyleSheet,
+}
+
+impl Renderer {
```
zbraniecki, line 16: `praise`: I like that abstraction. Great addition!

src/renderer/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+                emphasis: Style::new().effects(Effects::BOLD),
+                none: Style::new(),
+            },
+        }
+    }
+
+    pub const fn anonymized_line_numbers(mut self, anonymized_line_numbers: bool) -> Self {
+        self.anonymized_line_numbers = anonymized_line_numbers;
+        self
+    }
```
zbraniecki, line 56: `question`: Why do you need those setters, why not make the properties public?

zbraniecki (commented):
lgtm!

src/renderer/mod.rs
Muscraft: I did it because it is easier (and cleaner) to edit one specific field than making the fields public. The other thing is I believe it makes adding a field not a breaking change, but I do not know for certain.

zbraniecki: I don't understand how is it cleaner. Comparing:

```rust
let mut renderer = Renderer::plain();
renderer.anonymized_line_numbers(true);
```

vs

```rust
let mut renderer = Renderer::plain();
renderer.anonymized_line_numbers = true;
```

Can you explain further?

> The other thing is I believe it makes adding a field not a breaking change, but I do not know for certain.

That can be achieved via https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute

Muscraft: That's true I forgot about that. I will switch things over

Muscraft: I was looking at making them public, and I realized what I meant by cleaner.
```rust
let renderer = Renderer::plain()
    .anonymized_line_numbers(true)
    .margin(margin);
```

vs 

```rust
let mut renderer = Renderer::plain();
renderer.anonymized_line_numbers = true;
renderer.margin = margin;
```

It makes it so that `renderer` is not mutable and things can be chained together (omitting `renderer` each time).

It follows the builder pattern which I am a fan for smaller structs, for larger structs I do think making fields public is a good idea. 

If you would still like me to change it to be public fields I can but it will take me a moment to refactor.

epage: I wish you could use `..` initialization with `non_exhaustive`

zbraniecki: > It makes it so that renderer is not mutable and things can be chained together (omitting renderer each time).

That's not correct. The renderer still has to be mutable for the setters to work (you're consuming them btw. `mut self`).

The builder pattern is useful when your setters actually perform computations, invariant validation etc. In this case, your builder is the struct itself, so exposing it as such allows customers to use any Rust patterns to construct it.

By hiding it you're relying on custom uninspectable API to achieve the same.

It's a glorified version of:

```rust
struct Point {
    x: f64,
    y: f64
}

impl Point {
    fn new(x: f64, y: f64) -> Self {
        Self { x, y }    
    }

    fn set_x(&mut self, x: f64) -> &mut Self {
        self.x = x;
        self
    }

    fn set_y(&mut self, y: f64) -> &mut Self {
        self.y = y;
        self
    }
}
```

vs just:

```rust
struct Point {
    pub x: f64,
    pub y: f64,
}
```

and let customers create it and operate on it any way they want. You can add helper impl methods (like you do with `plain`), but taking away ability to operate on the fields without any reason seems like a way that limits your downstream consumers from being able to interact with the struct in ways that you don't necessarily have to be able to even predict.

In other words, if at the moment of consumption, you don't care how this struct has been constructred (maybe someone created their own methods to build the Renderer, or merge two), and you don't perform any validation, or variant locking, then just let it be an open struct and let people use all of Rust expressiveness to operate on it.

Saying that - this is my opinion, not a strong requirement - I will not try to block your PR on that, so if my arguments are not convincing to you, continue with this API and I'll r+ it

zbraniecki: You can: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6ddedead395636105b9830d26ab0e663

epage: Thats within the same visibility. I don't think you can if you don't have visibility.

One benefit I've seen to builder is the ability to evolve an API, particularly for evolving styles. You'll also see builder used in a lot of the ecosystem when no validation is done.

zbraniecki: Ah, good call. Yeah, ability to use `..` would be nice.

I'm generally supportive of builder patterns when there's logic involved. I find it inelegant when we have a plain struct and we populate it with dummy getters/setters. 

But as I stated, this is my preference, and I'm not intending to push for it any further. I support the direction the author of the PR decides to take.

zbraniecki (approved):
lgtm!

src/renderer/mod.rs
Muscraft: I can see the merits of both and could go either way on this. Even though it is a breaking change, we could always change in the future. I'd say we should go with the builder (for now).

src/lib.rs
```diff
@@ -26,29 +26,21 @@
...
-//! A user of the crate may also choose to provide their own formatter logic,
-//! to convert a `DisplayList` into a `String`, or just a `Stylesheet` to
-//! use the crate's formatting logic, but with a custom stylesheet.
-// TODO: check documentation
 
-pub mod display_list;
-pub mod formatter;
+mod display_list;
+pub mod renderer;
 pub mod snippet;
```
epage, line 46: Side note: I feel like too much of the public API is nested.  I can see moving all of `snippet` into the crate's root namespace and `Render`, leaving any other `Render` related types in `renderer`

src/renderer/margin.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl Margin {
+    pub fn new(
+        whitespace_left: usize,
+        span_left: usize,
+        span_right: usize,
+        label_right: usize,
+        column_width: usize,
+        max_line_len: usize,
+    ) -> Self {
```
epage, line 34: In the future, we might want to consider if there are improvements to this API as well

src/renderer/mod.rs
```diff
@@ -1,6 +1,9 @@
-use crate::display_list::{DisplayList, Margin};
+pub mod margin;
+
+use crate::display_list::DisplayList;
 use crate::snippet::Snippet;
 use anstyle::{AnsiColor, Effects, Style};
+use margin::Margin;
```
epage: We should `mod margin` and `pub use margin::Margin`

There is little gained from deeply nesting the API like this

src/renderer/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+//!  };
+//!
+//!  let renderer = Renderer::styled();
+//!  println!("{}", renderer.render(snippet));
+
+pub mod margin;
+
+use crate::display_list::DisplayList;
+use crate::snippet::Snippet;
+use anstyle::{AnsiColor, Effects, Style};
```
epage, line 39: note: we likely want to re-export `anstyle`

Some people literally re-export the crate in the root (`pub use anstyle;`).  Where possible, I prefer doing it in a way that makes sense from an API perspective.  For example, in clap we have [mod styling](https://docs.rs/clap/latest/clap/builder/styling/index.html) which has both clap's Stylesheeting and all of `anstyle`

src/renderer/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+impl Default for StyleSheet {
+    fn default() -> Self {
+        Self::const_default()
+    }
+}
+
+impl StyleSheet {
+    pub(crate) const fn const_default() -> Self {
```
epage, line 198: Why have this separate from `plain`?

When reading the code, its clearer whats intended when looking at the `iml Default` and seeing it call `Self::plain` than `Self::const_default`

epage: Whoops, I didn't pay attention to what type this was one.  In this case, I'd rename `const_default` to `plain`

src/renderer/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+            stylesheet: StyleSheet::const_default(),
+        }
+    }
+
+    /// Default terminal styling
+    pub const fn styled() -> Self {
+        Self {
+            anonymized_line_numbers: false,
+            margin: None,
+            stylesheet: StyleSheet {
```
epage, line 76: To make it clearer that some of these fields are meant to be the same as `plain`, should we do
```rust
Self {
    stylesheet: StyleSheet { ... },
    ..Self::plain(),
}
```

src/renderer/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct Renderer {
+    anonymized_line_numbers: bool,
+    margin: Option<Margin>,
+    stylesheet: StyleSheet,
+}
+
+impl Renderer {
+    /// Render a snippet into a `Display`able object
+    pub fn render<'a>(&'a self, snippet: Snippet<'a>) -> impl Display + 'a {
+        DisplayList::new(
```
epage, line 54: imo constructors should go first

src/renderer/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    /// Set the output style for none
+    pub const fn none(mut self, style: Style) -> Self {
+        self.stylesheet.none = style;
+        self
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct StyleSheet {
```
epage, line 180: nit: Should we move this out into a `mod`?

epage, line 180: nit: I tend to prefer `Stylesheet`

src/lib.rs
```diff
@@ -26,29 +26,25 @@
...
-//! to convert a `DisplayList` into a `String`, or just a `Stylesheet` to
-//! use the crate's formatting logic, but with a custom stylesheet.
-// TODO: check documentation
 
-pub mod display_list;
-pub mod formatter;
-pub mod snippet;
-pub mod stylesheets;
+mod display_list;
+pub mod renderer;
```
epage, line 45: At this point, `display_list` is an implementation detail of `renderer` and should likely go in there

epage (approved):
nit: refactors can't have breaking changes, so I'd change the commit types away from it where user-visible changes are made

PR #73, opened 2024-01-02
fix: Match rustc's colors

Muscraft:
This PR makes it so that we match `rustc`'s output regarding colors. See [here](https://github.com/rust-lang/rust/blob/e51e98dde6a60637b6a71b8105245b629ac3fe77/compiler/rustc_errors/src/emitter.rs#L2676-L2723) and [here](https://github.com/rust-lang/rust/blob/e51e98dde6a60637b6a71b8105245b629ac3fe77/compiler/rustc_errors/src/lib.rs#L1768-L1787) for `rustc`'s current color implementation.

Note: Where `rustc` uses `set_intense(true)`, that is the same as `Bright<color>` in `anstyle`.

Note: Most things are `BOLD` as `rustc` eventually sets most things bol (see the first link).

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -62,15 +62,30 @@ impl Renderer {
 
     /// Default terminal styling
     pub const fn styled() -> Self {
+        const BRIGHT_BLUE: Style = if cfg!(windows) {
+            AnsiColor::BrightCyan.on_default()
+        } else {
+            AnsiColor::BrightBlue.on_default()
+        };
         Self {
             stylesheet: Stylesheet {
                 error: AnsiColor::BrightRed.on_default().effects(Effects::BOLD),
-                warning: AnsiColor::BrightYellow.on_default().effects(Effects::BOLD),
-                info: AnsiColor::BrightBlue.on_default().effects(Effects::BOLD),
-                note: Style::new().effects(Effects::BOLD),
+                warning: if cfg!(windows) {
+                    AnsiColor::BrightYellow.on_default()
+                } else {
+                    AnsiColor::Yellow.on_default()
+                }
+                .effects(Effects::BOLD),
+                info: BRIGHT_BLUE.effects(Effects::BOLD),
+                note: AnsiColor::BrightGreen.on_default().effects(Effects::BOLD),
                 help: AnsiColor::BrightCyan.on_default().effects(Effects::BOLD),
-                line_no: AnsiColor::BrightBlue.on_default().effects(Effects::BOLD),
-                emphasis: Style::new().effects(Effects::BOLD),
+                line_no: BRIGHT_BLUE.effects(Effects::BOLD),
+                emphasis: if cfg!(windows) {
+                    AnsiColor::BrightWhite.on_default()
+                } else {
+                    Style::new()
+                }
+                .effects(Effects::BOLD),
                 none: Style::new(),
             },
             ..Self::plain()


src/renderer/mod.rs
```diff
@@ -62,15 +62,28 @@ impl Renderer {
 
     /// Default terminal styling
     pub const fn styled() -> Self {
+        const BRIGHT_BLUE: Style = if cfg!(windows) {
+            AnsiColor::BrightCyan.on_default().effects(Effects::BOLD)
+        } else {
+            AnsiColor::BrightBlue.on_default().effects(Effects::BOLD)
+        };
```
epage, line 69: nit: as this isn't just bright blue, lets move the `.effects` call down below

src/renderer/mod.rs
```diff
@@ -62,15 +62,28 @@ impl Renderer {
...
                 error: AnsiColor::BrightRed.on_default().effects(Effects::BOLD),
-                warning: AnsiColor::BrightYellow.on_default().effects(Effects::BOLD),
-                info: AnsiColor::BrightBlue.on_default().effects(Effects::BOLD),
-                note: Style::new().effects(Effects::BOLD),
+                warning: if cfg!(windows) {
+                    AnsiColor::BrightYellow.on_default().effects(Effects::BOLD)
+                } else {
+                    AnsiColor::Yellow.on_default().effects(Effects::BOLD)
+                },
+                info: BRIGHT_BLUE,
```
epage, line 78: What does `info` map to in `rustc`?  No level uses `BRIGHT_BLUE`.  Is it a Secondary?

src/renderer/mod.rs
```diff
@@ -62,15 +62,28 @@ impl Renderer {
...
         Self {
             stylesheet: Stylesheet {
                 error: AnsiColor::BrightRed.on_default().effects(Effects::BOLD),
-                warning: AnsiColor::BrightYellow.on_default().effects(Effects::BOLD),
-                info: AnsiColor::BrightBlue.on_default().effects(Effects::BOLD),
-                note: Style::new().effects(Effects::BOLD),
+                warning: if cfg!(windows) {
+                    AnsiColor::BrightYellow.on_default().effects(Effects::BOLD)
+                } else {
+                    AnsiColor::Yellow.on_default().effects(Effects::BOLD)
```
epage, line 76: nit: should we have `.effects` be outside of the `if` expression?

zbraniecki:
My slight preference would be to externalize platform-specific settings to some options bag parameter like "use_cyan" and set it in constructor to true for windows in the binary.

This makes it easier for a customer to replicate the same color scheme as windows has in other environments without having to emulate windows cfg.

src/renderer/mod.rs
```diff
@@ -62,15 +62,28 @@ impl Renderer {
...
+                },
+                info: BRIGHT_BLUE,
+                note: AnsiColor::BrightGreen.on_default().effects(Effects::BOLD),
                 help: AnsiColor::BrightCyan.on_default().effects(Effects::BOLD),
-                line_no: AnsiColor::BrightBlue.on_default().effects(Effects::BOLD),
-                emphasis: Style::new().effects(Effects::BOLD),
+                line_no: BRIGHT_BLUE,
+                emphasis: if cfg!(windows) {
```
epage, line 82: If I'm reading this correctly, this maps to `MainHeaderMsg`?  What about `Highlight`?

src/renderer/mod.rs
```diff
@@ -62,15 +62,28 @@ impl Renderer {
...
+                    AnsiColor::BrightWhite.on_default().effects(Effects::BOLD)
+                } else {
+                    Style::new().effects(Effects::BOLD)
+                },
                 none: Style::new(),
```
epage, line 87: nit: should `.effects` be on the outside of the `if`?

epage:
> My slight preference would be to externalize platform-specific settings to some options bag parameter like "use_cyan" and set it in constructor to true for windows in the binary.

This is all themable so they can do what they wish.

For myself, my preference is that the agreed-to cross-Rust-project behavior be the default here so we don't have to worry about what feature flags and constructor parameters we need to keep in sync

src/renderer/mod.rs
Muscraft: It should map to `UnderlineSecondary` or `LabelSecondary`, but `annotate-snippets` doesn't have the concept of primary/secondary, so we print `info: ` before it currently.

src/renderer/mod.rs
Muscraft: Correct, `emphasis` maps to `MainHeadingMsg`. As for `Highlight` `annotate-snippets` doesn't currently have anything for that concept (that I can tell).

Muscraft:
The goal for this change is to be as close to `rustc` as possible by default. Having to tell `annotate-snippets` that you need Windows-specific colors seems to step away from this idea, as a user would not be getting the same colors as `rustc` by default. This also makes the API slightly more complicated and would require a user to know that they need to call `use_cyan()` (or similar) on Windows to get the correct colors for their platform.

PR #74, opened 2024-01-04
fix: Allow highlighting one past end

Muscraft:
Being able to highlight one past the end of the given source if the source provided is from `.lines()` (and got its newline stripped) or is the last line in the file and you want to highlight the next "character". `cargo` needs something to do this, so it doesn't have to remember to add a new line in either scenario.

### DIFF
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -934,14 +934,15 @@ fn format_body(
 ) -> Vec<DisplayLine<'_>> {
     let source_len = slice.source.chars().count();
     if let Some(bigger) = slice.annotations.iter().find_map(|x| {
-        if source_len < x.range.1 {
+        // Allow highlighting one past the last character in the source.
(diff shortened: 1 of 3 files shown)


src/renderer/display_list.rs
```diff
@@ -934,14 +934,15 @@ fn format_body(
...
+        // Allow highlighting one past the last character in the source.
+        if source_len + 1 < x.range.1 {
             Some(x.range)
         } else {
             None
         }
     }) {
         panic!(
-            "SourceAnnotation range `{:?}` is bigger than source length `{}`",
+            "SourceAnnotation range `{:?}` is more than one bigger than source length `{}`",
```
epage, line 945:
```suggestion
            "SourceAnnotation range `{:?}` is beyond the end of buffer `{}`",
```

epage (approved):

PR #75, opened 2024-01-04
Annotate snippets 0.10.1

Muscraft:
This PR updates the `CHANGELOG.md` in preparation for `0.10.1`, as well as bumps the version number to `0.10.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+## [0.10.1] — January 4, 2024
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased] - ReleaseDate
 
+## [0.10.1] — January 4, 2024
+
+### Fixed
```
epage, line 11: If you have push permissions, you can use #76 and only need the changelog content and none of the version bumps.

Most of the time, I don't expect my changelogs to reviewed so I just have `cargo release` push it with the version bump, rather than creating a PR

Muscraft:
I used #76 and `cargo-release` instead of this PR

PR #82, opened 2024-02-29
feat: Add feature for testing colored output easier

Muscraft:
In [cargo#13461](https://github.com/rust-lang/cargo/pull/13461#discussion_r1501136340), it was noticed that the default colors being platform-specific made testing colored output quite challenging. This PR adds a feature that makes the colors the same across the board, which should be enabled when testing colored output.

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -61,8 +61,12 @@ impl Renderer {
     }
 
     /// Default terminal styling
+    ///
+    /// # Note
+    /// When testing styled terminal output, see the [`testing-colors` feature](crate#features)
     pub const fn styled() -> Self {
-        const BRIGHT_BLUE: Style = if cfg!(windows) {
+        const USE_WINDOWS_COLORS: bool = cfg!(windows) && !cfg!(feature = "testing-colors");
+        const BRIGHT_BLUE: Style = if USE_WINDOWS_COLORS {
             AnsiColor::BrightCyan.on_default()
         } else {
             AnsiColor::BrightBlue.on_default()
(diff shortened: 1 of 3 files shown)


src/renderer/mod.rs
```diff
@@ -60,7 +60,16 @@ impl Renderer {
         }
     }
 
+    #[cfg(not(feature = "testing-colors"))]
```
epage, line 63: I'd recommend having one public facing method
- Easier to avoid signature drift
- You only get the docs for one of them (and docs.rs matters the most) so you should have the documentation the same and doing that is the easiest when you have one public signature

src/renderer/mod.rs
```diff
@@ -60,7 +60,16 @@ impl Renderer {
...
+    ///
+    /// # Note
+    /// Some colors are OS dependant, meaning that the colors will look
+    /// different on different operating systems, making testing colored output
+    /// difficult.
+    /// If you are testing the colored output, you can use the
+    /// [`testing-colors` feature](crate#features) to enable colors that are OS
+    /// independent.
     pub const fn styled() -> Self {
         const BRIGHT_BLUE: Style = if cfg!(windows) {
```
epage, line 74: Should we try to single-source these?  If nothing else, we could have an internal `const fn` that has a `windows: bool` parameter.

src/renderer/mod.rs
```diff
@@ -60,7 +60,16 @@ impl Renderer {
...
+    #[cfg(not(feature = "testing-colors"))]
     /// Default terminal styling
+    ///
+    /// # Note
+    /// Some colors are OS dependant, meaning that the colors will look
+    /// different on different operating systems, making testing colored output
+    /// difficult.
+    /// If you are testing the colored output, you can use the
+    /// [`testing-colors` feature](crate#features) to enable colors that are OS
+    /// independent.
```
epage, line 72: I'd explicitly call this out as a `dev-dependency`, maybe showing the command `cargo add annotate-snippets --dev --feature testing-colors`

PR #86, opened 2024-03-02
refactor(fixtures): Move to snapbox with SVGs

Muscraft:
I was not happy with how fixtures were currently being run, as when one failed it would take down the whole thing failed, and it could not continue past the failed fixture test. To make this better I switched the fixture tests to use a custom test harness, built on top of [`snapbox::harness::Harness`](https://docs.rs/snapbox/latest/snapbox/harness/struct.Harness.html). This makes it so each test is run by itself so you can see exactly which test is failing. In doing this I also switched the expected output to be compared to SVGs, which makes seeing visual changes much easier, while still retaining the ability to see the text differences.


tests/fixtures/main.rs
```diff
@@ -0,0 +1,33 @@
+mod deserialize;
+
+use crate::deserialize::Fixture;
+use annotate_snippets::{Renderer, Snippet};
+use snapbox::data::DataFormat;
+use snapbox::Data;
+use std::error::Error;
+
+fn main() {
+    #[cfg(not(windows))]
```
epage, line 10: Why `not(windows)`?

Muscraft: There are deserialization errors on Windows. This was a problem with the [previous fixture test](https://github.com/rust-lang/annotate-snippets-rs/blob/e9b445c3d43dc5d1673a89dcdab691187a6b9566/tests/fixtures_test.rs#L22).

Cargo.toml
```diff
@@ -28,16 +28,22 @@ anstyle = "1.0.4"
 unicode-width = "0.1.11"
 
 [dev-dependencies]
+annotate-snippets = { path = ".", features = ["testing-colors"] }
```
epage, line 31: With this you can't test the platform-specific colors in annotate-snippets

Some options
- Offer a "test" constructor used by most tests except for the couple that need platform-specific colors
- Have a test focused solely on the colors (no annotations) that outputs a line per color (with text saying what the semantic meaning is).
  - Something like this is my preferred route as it makes the platform-specific testing (the most annoying to update) be independent of most of the rest of the styling so you can make changes in the output and not have to update platform-specific snapshots

Muscraft: I realized this doesn't need to be set here, and should probably be done in a separate PR, when we are actually testing colors

Issue #77, opened 2024-01-10
Source annotate ranges should use bytes not chars


Muscraft:
Currently, `SourceAnnotation`'s `range` field is based on chars, not bytes. This can be a little confusing for users and can cause issues when dealing with spans from things like `toml` errors. It would make more sense to have users give us bytes, and we calculate the char spans internally. This would make the API easier to use as most people can handle going to bytes, but going from bytes -> chars can be very hard.

epage:
For me, there are two aspects to this
- `annotate-snippets` is responsible for rendering so it shouldn't accept a rendering-focused unit but handle that behind the scenes
- Turning bytes into column counts is complicated and we likely don't want everyone re-implementing it incorrectly, in different ways

PR #90, opened 2024-03-08
fix!: Take in byte spans

Muscraft:
BREAKING CHANGE: This switches from char spans to byte spans

This PR is a draft implementation to address #77. While I think I fixed every issue in moving from char spans to byte spans, I am not entirely certain. I may look into adding more tests to ensure we are doing the write thing. This also doesn't address the fact that people could pass in char spans when updating the version, not realizing the change. We could do something to make it more apparent but I am not sure what that would be.

### DIFF
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -32,6 +32,8 @@
 //!
 //! The above snippet has been built out of the following structure:
 use crate::snippet;
+use itertools::FoldWhile::{Continue, Done};
+use itertools::Itertools;
 use std::fmt::{Display, Write};
 use std::{cmp, fmt};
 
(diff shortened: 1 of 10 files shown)


src/renderer/display_list.rs
```diff
@@ -932,7 +948,7 @@ fn format_body(
     has_footer: bool,
     margin: Option<Margin>,
 ) -> Vec<DisplayLine<'_>> {
-    let source_len = slice.source.chars().count();
+    let source_len = slice.source.as_bytes().iter().count();
```
epage, line 951: Isn't this just `slice.source.len()`?

epage (approved):

PR #91, opened 2024-03-09
feat!: Move to builder pattern for snippet creation

### LINKED ISSUE #87: Make API more resilient to change by using Builder Pattern
Currently, the API directly initializes structs.  This makes it harder to evolve the API.  We should switch to the builder pattern for creating snippets.

Muscraft:
BREAKING CHANGE: Snippets must be created using the builder pattern

This PR makes it so that `Snippet`s must be created via a builder pattern, as described in #87. This should make it easier to change the internal API without making breaking changes. It also appears to help in code readability and size but that is determined by personal preference.

fixes #87

### DIFF
--- a/src/snippet.rs
+++ b/src/snippet.rs
@@ -5,40 +5,121 @@
 //! ```
 //! use annotate_snippets::*;
 //!
-//! Snippet {
-//!     title: Some(Annotation {
-//!         label: Some("mismatched types"),
-//!         id: None,
-//!         annotation_type: AnnotationType::Error,
-//!     }),
-//!     footer: vec![],
-//!     slices: vec![
-//!         Slice {
-//!             source: "Foo",
-//!             line_start: 51,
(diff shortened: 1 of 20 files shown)


src/snippet.rs
```diff
@@ -5,56 +5,201 @@
...
+        }
+    }
+
+    pub fn label(mut self, label: &'a str) -> Self {
+        self.label = label;
+        self
+    }
+
+    /// Create a [SourceAnnotation] with the given span for a [Slice]
+    pub fn span(&self, span: Range<usize>) -> SourceAnnotation<'a> {
```
epage, line 147: We could generalize this to any range type. up to you

epage (approved):

src/snippet.rs
Muscraft: I think this is something we could do later if asked for it

PR #107, opened 2024-03-21
fix: Allow empty annotation spans

epage:
Cargo works with empty spans in some situations, especially for eol/eof.  Annotate-snippets had a test case for this but it was dependent on a hack that remained in Cargo.  This unblocks Cargo from removing that hack.  I tested my local cargo against this and it fixes the problem I had there.

In the future, we may want to consider using a special form of underlining for empty spans.


Muscraft (approved):
Thanks!

If you want to create an issue for "a special form of underlining for empty span" that would be great!

PR #109, opened 2024-03-21
fix(render): On fold, strip prefix/suffix rather than fold 

epage:
As we don't show `...` on non-first lines without fold, it seems to make
sense to do the same thing for fold.
To show this, we likely would want to add more nuance than just `fold`.

This has the benefit of offering a "magic mode" where the annotated
lines get selected automatically.

This was implemented by pre-processing things which should keep overhead
low as other calculations down the line aren't done.


Muscraft (approved):
I think this is a good change overall, but we will want to add the ability for a user to specify the number of lines to keep.

PR #111, opened 2024-03-21
fix(render): Fold like rustc

epage:
Before, we elided all small sets of unhighlighted lines and then partially elided large sets of unhighlighted lines (favoring the start).

In contrast, rustc shows 2 lines of context, split between the start and end of the fold.
If there are 3 unhighlighted lines, it just skips folding. This makes a lot more sense.
See https://github.com/rust-lang/rust/blob/2627e9f3012a97d3136b3e11bf6bd0853c38a534/compiler/rustc_errors/src/emitter.rs#L1878C17-L1939C18

In contrast to rustc, I made the amount of context a variable to make the intent clearer and to open the door to it being configurable.


Muscraft (approved):
Thanks!

PR #124, opened 2024-06-06
chore: Update from epage/_rust template

epage:
- Use `cargo hack` for testing feature combinations
- Use `cargo hack` for lower overhead MSRV updates
- Switches to manual MSRV update (since overhead is so low)
- Pin the compiler for the clippy job so CI is always green
- Expand the lints used by default
- Resolve `cargo deny` deprecations
- Optimize CI by auto-canceling jobs
- Have a parent job for reporting CI status

Similar updates:
- winnow-rs/winnow#515
- clap-rs/clap#5492
- assert-rs/snapbox#304


Muscraft (approved):
Thanks!

PR #126, opened 2024-06-06
chore: Drop MSRV to 1.65

epage:
With the upcoming MSRV-aware resolver, I've been experimenting with lower MSRVs.

This will directly benefit `winnow` in allowing it to have examples of having parse errors rendered with annotate-snippets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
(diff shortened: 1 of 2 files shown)


Muscraft (approved):
I am all for this change for a few reasons:
- We have had [comments in the past](https://github.com/rust-lang/annotate-snippets-rs/pull/71#issuecomment-1886665129) asking why we updated the MSRV
- We can support more crates (I am very happy to hear `winnow` is considering using `annotate-snippets`)
- It doesn't look like it changes many dependency versions

PR #133, opened 2024-06-19
Match Rust's multiline annotation output 

Muscraft:
One big difference between Rust's output and `annotate-snippets` is that `annotate-snippets` did not work well when multiple annotations were on a single line. To show this, I added tests for multiple annotations per line and tests from [Rust's parser tests](https://github.com/rust-lang/rust/blob/894f7a4ba6554d3797404bbf550d9919df060b97/compiler/rustc_parse/src/parser/tests.rs). Then, in the next few commits I worked to match Rust's output.

Note: Most of the code to make this work was adapted from [Rust's `Emitter`](https://github.com/rust-lang/rust/blob/894f7a4ba6554d3797404bbf550d9919df060b97/compiler/rustc_errors/src/emitter.rs#L706-L1155). This has the added benefit of making our output much closer to Rust's.


tests/formatter.rs
```diff
@@ -1,3 +1,9 @@
+//! Tests for the output format of the renderer.
+//!
+//! Tests that start with `rustc` have been adapted from [Rust's parser tests][parser-tests].
+//!
+//! [parser-tests]: https://github.com/rust-lang/rust/blob/894f7a4ba6554d3797404bbf550d9919df060b97/compiler/rustc_parse/src/parser/tests.rs
```
epage, line 5: Maybe we should split these out into a separate test file?

src/snippet.rs
```diff
@@ -108,7 +108,7 @@ impl<'a> Snippet<'a> {
 /// An annotation for a [`Snippet`].
 ///
 /// See [`Level::span`] to create a [`Annotation`]
-#[derive(Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
epage, line 111: Should likely split these API level changes out into their own commit with the motivations

tests/fixtures/no-color/multiple_annotations.svg
```diff
@@ -20,15 +20,15 @@
 </tspan>
     <tspan x="10px" y="46px"><tspan>    |</tspan>
 </tspan>
-    <tspan x="10px" y="64px"><tspan> 96 | fn add_title_line(result: &amp;mut Vec&lt;String&gt;, main_annotation: Option&lt;&amp;Annotation&gt;) {</tspan>
+    <tspan x="10px" y="64px"><tspan>96  | fn add_title_line(result: &amp;mut Vec&lt;String&gt;, main_annotation: Option&lt;&amp;Annotation&gt;) {</tspan>
```
epage, line 23: rustc left-aligns numbers?  That is an odd choice and want to make sure we make this choice knowingly (and considering whether we should deviate)

epage: This is also a change that is separate from the commit title / PR

examples/expected_type.svg
```diff
@@ -19,25 +19,25 @@
   <rect width="100%" height="100%" y="0" rx="4.5" class="bg" />
 
   <text xml:space="preserve" class="container fg">
-    <tspan x="10px" y="28px"><tspan class="fg-bright-red bold">error</tspan><tspan>: </tspan><tspan class="bold">expected type, found `22`</tspan>
+    <tspan x="10px" y="28px"><tspan class="fg-bright-red bold">error</tspan><tspan class="bold">: expected type, found `22`</tspan>
```
epage, line 22: While reasonable, this is another change that was not called out in the PR / commit title

tests/formatter.rs
```diff
@@ -795,8 +801,9 @@ bar = { version = "0.1.0", optional = true }
 error: unused optional dependency
   |
 4 |   bar = { version = "0.1.0", optional = true }
-  |  __________________________________________^
-  |                              --------------- info: This should also be long but not too long
+  |  ____________________________--------------^
+  | |                            |
+  | |                            info: This should also be long but not too long
```
epage, line 806: Am I reading this right that the annotation spans overlap but the rendering is not showing that?

Also, whats with the trailing caret?

tests/formatter.rs
```diff
@@ -837,14 +844,16 @@ bar = { version = "0.1.0", optional = true }
...
+  | ||
 5 | || this is another line
 6 | || so is this
 7 | || bar = { version = "0.1.0", optional = true }
-  | ||__________________________________________^ I need this to be really long so I can test overlaps
-  | ||_________________________^ I need this to be really long so I can test overlaps
+  | ||_________________________^________________^ I need this to be really long so I can test overlaps
+  | |__________________________|
+  |                            I need this to be really long so I can test overlaps
   |
```
epage, line 857: Is the diff being funky or is this really confusing to see the spans highlighted?

src/snippet.rs
Muscraft: I was able to remove them entirely

tests/fixtures/no-color/multiple_annotations.svg
Muscraft: They were left-aligned in the tests I ported. Either way, I removed that from this PR and will create a PR just for it

examples/expected_type.svg
Muscraft: I think this was a relic from some other changes, so I removed it

tests/formatter.rs
Muscraft: Yes, they do overlap, but it is being shown. Multiline annotation underlines get written first, then any `-` and `^`. In this case it is showing they overlap with the `-` intersecting the multiline annotation.

tests/formatter.rs
Muscraft: Maybe I have the curse of knowledge, but it makes sense to me; what part is confusing?

tests/rustc_tests.rs
```diff
@@ -0,0 +1,761 @@
...
+    let expected = str![[r#"
+error: foo
+ --> test.rs:3:7
+  |
+3 |   a { b { c } d }
+  |   ^^^^^^^^^^^^^
+  |       |
+  |       `b` is a good letter
+  |
+"#]];
```
estebank, line 452: For this case, you really want to have different label levels, so that you can confirm that overlapping spans of different levels do not "hide" each other. In `rustc` we ensure we print from widest to shortest span, so that this ends up like `^^^^-----^^^^` or `----^^^^----`. This isn't *perfect* but it is the best we can do when restricted to ASCII and a single line of text.

tests/rustc_tests.rs
```diff
@@ -0,0 +1,761 @@
...
+    // This is all `^` until we support primary and secondary labels
+    let expected = str![[r#"
+error: foo
+ --> test.rs:3:6
+  |
+3 |      X0 Y0 Z0
+  |  _______^
+4 | |    X1 Y1 Z1
+  | | ____^_^
+  | ||____|
```
estebank, line 281: This is something I want to handle in rustc eventually: when we have only multiline spans pointing at a line of code, invert the order of the vertical lines so that there are no overlaps anymore. (Future work/Not a blocker)

tests/formatter.rs
estebank: The diff makes it harder to read, but the new output matches rustc's

tests/rustc_tests.rs
Muscraft: This was added to Rust in rust-lang/rust#126882, and added to this PR in the latest commit

estebank (approved):

tests/rustc_tests.rs
epage: ... yeah, I think my comment from the other issue stands.  I think rustc is being too "smart" and a lot of these overlapping spans and weird cases are bonkers for a user to decipher

epage (approved):
To be blunt, I think this PR is a major regression and that rustc is in the wrong to render overlapping spans.  However, rustc is the "gold standard" and matching it is the goal, so I'm reluctantly fine with this moving forward.

estebank:
@epage I think it is reasonable for annotate-snippets to  (eventually) allow configuring this (as other tools might want different behavior).

For the future, I'd want something more bold: being able to emit SVG output with much fancier output that can cram multiple different separated underlines in the space a single line of text would fit, and as an even bigger stretch, use sixels for rendering the underlines. But none of these things are reasonable to attempt 1) now or 2) in rustc first.

epage:
Configuration breeds complexity.  Just for what is in this PR, I don't think configuration is worth it.

Issue #110, opened 2024-03-21
Allow configuring context around `fold`ed code


epage:
#109 fixed an inconsistency in how we folded code before and after all of the annotations.

Callers would likely want to control how much context is shown around each annotation.

chengr4:
Hi @epage,

I’m also working on addressing this ticket and would appreciate your guidance again lol.

---

My imagination is to create a new function eg `show_context(num: u8)` which works when `snippet.fold == false` for `Snippet`.

> ps. in the function, counting Max context line for each annotation is needed.

epage:
It'd likely be best for us to hold off on an issue like this until we have people who specifically need it so we can gather requirements rather than trying to guess what users need.

epage:
One idea we've discussed for configuring context is the ability to force folded content to be shown.  We could do this by a new `AnnotationKind`.  For example, you could annotate a function name and the close symbol for a function with this new `AnnotationKind` to provide extra context for where an error occurred.  This seems much richer than customizing N lines before and M lines after.  One question is how this should interact with the code wider than the screen when the edges get cut off.  Should we force the annotated section to be visible, or only the line for that section to be visible?

PR #140, opened 2024-08-14
Add unit test for function fold_prefix_suffix

chengr4:
As I was trying to understand #109, #110, I took the opportunity to add unit tests. I hope I'm on the right way.

## References

- #109 
- #110


epage:
The renderer has changed dramatically and we have end-to-end tests for folding, so I'm going to go ahead and close this.

Issue #101, opened 2024-03-13
Support any `Display`, not just `&str`


epage:
This is split out of https://github.com/rust-lang/annotate-snippets-rs/issues/13#issuecomment-543321203

> Input text doesn't have to be &str; it can be anything that can implement Display. This is especially useful when the origin source https://github.com/rust-lang/annotate-snippets-rs/pull/12#discussion_r322920987, so the intermediate allocation doesn't have to happen. Playing with take-two I think this dyn has negligible effect on runtime, the real killer to my time is dyn indirection to the span resolvers IIUC.

epage:
Since we'd have to take a `&dyn Display` and `to_string()` it anyways, I feel this isn't offering us much and am going to go ahead and close this.  If there is a reason we should re-visit this, let us know!

PR #141, opened 2024-08-19
docs: Fix and add docs for `Level`

icecream17:
Have the documentation for `Level` reflect actual appearance.

---

The output from this file:

<details><summary>(43 lines)</summary>

```rust
//! Provides a user-friendly error if the given file has the string `tab`
//!
//! A silly utility to test annotate_snippets (and anstream)

use clap::Parser;

#[derive(Parser)]
struct Cli {
    file_path: std::path::PathBuf,
}

fn main() {
    let Cli { file_path } = Cli::parse();
    let path_string = file_path.to_string_lossy().to_string();

    // Bad error message; don't do this!
    let contents =
        std::fs::read_to_string(file_path).expect("Should have been able to read the file");

    let message = if let Some(index) = contents.find("tab") {
        Level::Error.title("no tabs allowed!").snippet(
            Snippet::source(&contents)
                .line_start(1)
                .origin(&path_string)
                .fold(true)
                .annotation(
                    Level::Error
                        .span(index..(index + 3))
                        .label("Found `tab` here"),
                )
                .annotation(Level::Warning.span(index..(index + 3)).label("Warning"))
                .annotation(Level::Info.span(index..(index + 3)).label("Info"))
                .annotation(Level::Note.span(index..(index + 3)).label("Note"))
                .annotation(Level::Help.span(index..(index + 3)).label("remove `tab`")),
        )
    } else {
        Level::Info.title("No \"tabs\" found, congrats!")
    };
    anstream::println!("{}", renderer.render(message));
}
```

</details>

Results in this output:
image

### DIFF
--- a/src/snippet.rs
+++ b/src/snippet.rs
@@ -126,12 +126,15 @@ impl<'a> Annotation<'a> {
 /// Types of annotations.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Level {
-    /// Error annotations are displayed using red color and "^" character.
+    /// Error annotations are displayed in red with "^" characters.
     Error,
-    /// Warning annotations are displayed using blue color and "-" character.
+    /// Warning annotations are displayed in yellow with "-" characters.
     Warning,
+    /// Info annotations are displayed in blue with "-" characters and the prefix `info: `.
     Info,
+    /// Note annotations are displayed in green with "-" characters and the prefix `note: `.
     Note,
+    /// Help annotations are displayed in blue with "-" characters and the prefix `help: `.
     Help,
 }
 


icecream17:
Note: this is potentially inconsistent with cargo and clippy
image
image

icecream17:
Note: This is cyan rather than blue:
(found by using manual output while trying to work around #101:)
image

If a reviewer comments, I will change the wording to "cyan"

Muscraft:
Thank you for wanting to make them consistent.  However, I think the original doc comments mentioning the color and character was not the right call. This is primarily due to three reasons:
1. Colors are configurable by end users
2. Colors are platform dependent
3. Characters being tied to a `Level` is for historical reasons and will be removed in the future when we match `rustc`'s idea of `Primary` and `Secondary` for `Annotation`s

I do think it is a good idea to document the `Level` text and default color, but graphically and in a way that won't go stale. This can be done by utilizing [`fixture` tests](https://github.com/rust-lang/annotate-snippets-rs/tree/master/tests/fixtures) and [including the SVGs as doc comments](https://github.com/rust-lang/annotate-snippets-rs/blob/43a10aa8619972e446213a1c8a6649893b082f4b/src/lib.rs#L12). The rough process would be:
- Add a new `docs` folder inside of `tests/fixtures` for the new tests
- Add a basic example for each `Level`
- Link to the newly created SVGs in the [same way as we do in the lib docs](https://github.com/rust-lang/annotate-snippets-rs/blob/43a10aa8619972e446213a1c8a6649893b082f4b/src/lib.rs#L12)

PR #142, opened 2024-08-20
Fix line number 0 unaligned display

### LINKED ISSUE #57: Unaligned display when one line source given
Result:

```
 |
0 | SELECT bar
 |        ^^^ unexpected token
 | ^^^^^^^^^^ while parsing statement
 |
```

As you can see the `|` is not aligned.

Code:

```rust
pub(crate) fn display_annotated_error(source: &str, labels: Vec<(Range, String)>) -> String {
    let annotations = labels
        .iter()
        .map(|label| SourceAnnotation {
            range: (label.0.start, label.0.end),
            label: label.1.as_str(),
        })
        .collect::<Vec<_>>();

    let snippet = Snippet {
        title: None,
        footer: vec![],
        slices: vec![Slice {
            source,
            line_start: 0,
            origin: None,
            annotations,
            fold: false,
        }],
        opt: FormatOptions {
            color: true,
            ..Default::default()
        },
    };

    format!("{}", DisplayList::from(snippet))
}

display_annotated_error("SELECT bar", labels)
// where labels = [((7, 10), unexpected token), ((0, 10), while parsing statement)],
```

chengr4:
fix #57

### DIFF
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -72,12 +72,7 @@ impl<'a> fmt::Debug for DisplayList<'a> {
 
 impl<'a> Display for DisplayList<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let max_lineno = self.body.iter().fold(None, |max, set| {
-            set.display_lines.iter().fold(max, |max, line| match line {
-                DisplayLine::Source { lineno, .. } => std::cmp::max(max, *lineno),
-                _ => max,
-            })
-        });
+        let max_lineno = get_max_lineno(&self.body);
         let lineno_width = match max_lineno {
             None => 0,
             Some(_max) if self.anonymized_line_numbers => ANONYMIZED_LINE_NUM.len(),
@@ -147,6 +142,16 @@ impl<'a> DisplayList<'a> {
         }
         Ok(())
     }
+}
+
+fn get_max_lineno(body: &[DisplaySet<'_>]) -> Option<usize> {
+    let max_lineno = body.iter().fold(None, |max, set| {
+        set.display_lines.iter().fold(max, |max, line| match line {
+            DisplayLine::Source { lineno, .. } => cmp::max(max, *lineno),
+            _ => max,
+        })
+    });
+    max_lineno
 }
 
 #[derive(Debug, PartialEq)]
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -72,18 +72,17 @@ impl<'a> fmt::Debug for DisplayList<'a> {
 
 impl<'a> Display for DisplayList<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let lineno_width = self.body.iter().fold(0, |max, set| {
+        let lineno_width = self.body.iter().fold(None, |max, set| {
             set.display_lines.iter().fold(max, |max, line| match line {
-                DisplayLine::Source { lineno, .. } => cmp::max(lineno.unwrap_or(0), max),
+                DisplayLine::Source { lineno, .. } => std::cmp::max(max, *lineno),
                 _ => max,
             })
         });
-        let lineno_width = if lineno_width == 0 {
-            lineno_width
-        } else if self.anonymized_line_numbers {
-            ANONYMIZED_LINE_NUM.len()
-        } else {
-            ((lineno_width as f64).log10().floor() as usize) + 1
+        let lineno_width = match lineno_width {
+            None => 0,
+            Some(_max) if self.anonymized_line_numbers => ANONYMIZED_LINE_NUM.len(),
+            Some(0) => 1,
+            Some(max) => (max as f64).log10().floor() as usize + 1,
         };
 
         let multiline_depth = self.body.iter().fold(0, |max, set| {
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -72,13 +72,13 @@ impl<'a> fmt::Debug for DisplayList<'a> {
 
 impl<'a> Display for DisplayList<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let lineno_width = self.body.iter().fold(None, |max, set| {
+        let max_lineno = self.body.iter().fold(None, |max, set| {
             set.display_lines.iter().fold(max, |max, line| match line {
                 DisplayLine::Source { lineno, .. } => std::cmp::max(max, *lineno),
                 _ => max,
             })
         });
-        let lineno_width = match lineno_width {
+        let lineno_width = match max_lineno {
             None => 0,
             Some(_max) if self.anonymized_line_numbers => ANONYMIZED_LINE_NUM.len(),
             Some(0) => 1,
--- a/tests/formatter.rs
+++ b/tests/formatter.rs
@@ -905,3 +905,25 @@ error: unused optional dependency
     let renderer = Renderer::plain();
     assert_data_eq!(renderer.render(input).to_string(), expected);
 }
+
+// for issue 57
+#[test]
+fn test_line_number_0() {
+    let input = Level::Error.title("dummy").snippet(
+        Snippet::source("foo")
+            .origin("file/path")
+            .line_start(0)
+            .annotation(Level::Error.span(2..3)), // bar\nbaz
+    );
+
+    let expected = str![[r#"
+error: dummy
+ --> file/path:0:3
+  |
+0 | foo
+  |   ^
+  |
+"#]];
+    let renderer = Renderer::plain();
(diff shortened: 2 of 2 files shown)


tests/formatter.rs
```diff
@@ -279,9 +279,9 @@ fn test_only_source() {
         .snippet(Snippet::source("").origin("file.rs"));
     let expected = str![[r#"
 error
---> file.rs
- |
- |
+ --> file.rs
+  |
+  |
```
epage, line 284: I wonder if we shouldn't affect this case.  Maybe we should track `max_lineno` as an `Option` so `Some(0)` can have a width of 1 while `None` can have a width of 0.

chengr4: I will explain my train of thought at #57  to help you estimate it.

chengr4: @epage I have written down my train of thought [here](https://github.com/rust-lang/annotate-snippets-rs/issues/57#issuecomment-2301268251).

In fact, I don't quite understand why `max_lineno` might be `None`. I assume it would definitely be `usize` in my mind.

chengr4:
BTW, I was thinking to add unit tests for eg. the `format_line` method. it seemed quite challenging 😅, so I gave up. If anyone is willing to guide me on this, I would be very happy to give it another try.

tests/formatter.rs
epage: The answer to that question I leave to @Muscraft and @estebank

estebank: I see. I think that you should check for `self.body.iter().any(|set| set.display_lines.len() > 0)`. If `false`, then you only have a message and file path, so you don't *need* to move right. That case in rustc is rendered:

```
error: message
---> file.rs
```

without any attempt to print the `|`s. It seems to me that there would be left-over work to match that behavior.

That situation can occur when we have a span, but the underlying code couldn't be accessed for printing (this can happen with foreign crates or remapped paths).

All that to say, I think the previous output was better behavior, but I won't block either way.

chengr4: Now I understand what both of you are trying to tell me. It is true that `max_lineno` could be `None`! Thank you.

I've made an update. Please take a look when you have time 🙏.

chengr4:
FYI 
In the last commit, I extracted the calculation to another method. I'm not sure if it is proper, but I'm trying to make `fmt` adhere more closely to the Single Responsibility Principle

tests/formatter.rs
```diff
@@ -905,3 +905,25 @@ error: unused optional dependency
     let renderer = Renderer::plain();
     assert_data_eq!(renderer.render(input).to_string(), expected);
 }
+
+// for issue 57
+#[test]
+fn test_line_number_0() {
```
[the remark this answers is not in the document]
epage: > For someone viewing the commit with the behavior change, it makes it obvious what the intended behavior change is

This means the commit that changes behavior should also include test updates. Every commit should pass tests

src/renderer/display_list.rs
```diff
@@ -72,18 +72,22 @@ impl<'a> fmt::Debug for DisplayList<'a> {
 
 impl<'a> Display for DisplayList<'a> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let max_lineno = self.body.iter().fold(0, |max, set| {
+        let max_lineno = self.body.iter().fold(None, |max, set| {
```
epage: Feel free to squash this down into the other fix

src/renderer/display_list.rs
```diff
@@ -72,18 +72,22 @@ impl<'a> fmt::Debug for DisplayList<'a> {
...
             set.display_lines.iter().fold(max, |max, line| match line {
-                DisplayLine::Source { lineno, .. } => cmp::max(lineno.unwrap_or(0), max),
+                DisplayLine::Source { lineno, .. } => match (max, lineno) {
+                    (Some(max), Some(lineno)) => Some(cmp::max(max, *lineno)),
+                    (Some(max), None) => Some(max),
+                    (None, Some(lineno)) => Some(*lineno),
+                    (None, None) => None,
```
epage: You can do `std::cmp::max(max, lineno)` without a match.  `Some` compares higher than `None`

src/renderer/display_list.rs
```diff
@@ -152,6 +136,27 @@ impl<'a> DisplayList<'a> {
         }
         Ok(())
     }
+
+    fn get_lineno_width(&self) -> usize {
```
epage, line 140: imo I would instead make this `get_max_lineno()` which keep this focused and ensure policy around `anonymized_line_numbers` is obvious

chengr4: Just want to confirm you wish it to become

1.

```rust
fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    let max_lineno = self.get_max_lineno();
    let lineno_width = match max_lineno {
        None => 0,
        Some(_max) if self.anonymized_line_numbers => ANONYMIZED_LINE_NUM.len(),
        Some(0) => 1,
        Some(max) => (max as f64).log10().floor() as usize + 1,
   };

    // ...
}
```

2.

```rust
fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    let lineno_width = self.get_max_lineno();
    
    // ...
}
```
3.

```rust
fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    let max_lineno = self.get_max_lineno();
    let lineno_width = get_lineno_width(max_lineno);
    // ...
}
```

src/renderer/display_list.rs
```diff
@@ -148,6 +136,27 @@
         }
         Ok(())
     }
+
+    fn get_max_lineno(&self) -> Option<usize> {
+        let max_lineno = self.body.iter().fold(None, |max, set| {
+            set.display_lines.iter().fold(max, |max, line| match line {
+                DisplayLine::Source { lineno, .. } => std::cmp::max(max, *lineno),
```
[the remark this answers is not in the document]
chengr4: 😅 😳

chengr4:
@epage 
I thought about it and decided to refactor like this. If you feel the last commit is over-engineering, please let me know, and I'll delete it. Thanks.

src/renderer/display_list.rs
```diff
@@ -147,6 +142,16 @@ impl<'a> DisplayList<'a> {
         }
         Ok(())
     }
+
+    fn get_max_lineno(&self) -> Option<usize> {
+        let max_lineno = self.body.iter().fold(None, |max, set| {
```
epage, line 141: As this only needs state for getting `self.body`, let's make this a free function (ie not in an `impl` block and not taking `&self`)

epage:
> I thought about it and decided to refactor like this. If you feel the last commit is over-engineering, please let me know, and I'll delete it. Thanks.

For myself, I don't think its needed.

btw I've not been looking too deeply at this until the commits get cleaned up

src/renderer/display_list.rs
chengr4: I don't understand why a free function would be better than a method in this case, but I've made it

epage:
> btw I've not been looking too deeply at this until the commits get cleaned up

Please clean up the commits to be how you want them reviewed and merged.

I would recommend
- refactor commits
- test demonstrating the problem
- fix with test update showing new behavior

tests/formatter.rs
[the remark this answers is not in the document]
epage: As reminder, every commit should pass tests

src/renderer/display_list.rs
```diff
@@ -79,7 +79,7 @@ impl<'a> Display for DisplayList<'a> {
             })
         });
         let lineno_width = if lineno_width == 0 {
-            lineno_width
+            1
```
epage: I would recommend removing this commit

chengr4: fixed

tests/formatter.rs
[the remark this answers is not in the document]
epage: ae4694f84d6fcd2ab5bce4d05e0ae68997514326 is said to be a fix and changes behavior but there is no test update in that commit.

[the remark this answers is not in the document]
chengr4: Sorry I did not see the comment and please check out the update. Thanks.

chengr4:
@epage I know this PR isn't a very necessary fix, but I'm curious if we can continue with it or if you'd prefer to close it?

epage:
Looks like the commits aren't atomic. The "fix" commit doesn't include any test changes.

Note when i asked for a test commit, I asked that it show the existing behavior (it should pass) and that the fix commit then makes the test to work with the fix. This makes it very clear what the change in behavior is.

chengr4:
@epage I was thinking that these "test results" changes might be the 'test updates' you were hoping to see. If it is still incorrect, please guide me again. Thank you 🥹

epage:
As I said, this appears to be fixed already. I'm assuming it's from swapping out the entire renderer recently for rustc's. Look forward to future contributions though!

PR #144, opened 2024-08-21
Show fixture test location

Muscraft:
While working on #143, I noticed that fixture test names were missing the "module" (folder) before the name, which usually gets added for normal tests in folders. While this isn't a big deal now, if other test folders get added in the future, it will make it easier to find the failure.

Before:
Screenshot from 2024-08-21 15-31-07

After:
Screenshot from 2024-08-21 16-44-56


epage:
imo showing the source path is more helpful as it helps people more quickly connect with what action they can take and the can copy/paste.

PR #147, opened 2024-09-27
chore: Adopt '_rust/main' template

epage:
I needed to update the pre-commit Action and decided this would make it easier

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - uses: taiki-e/install-action@cargo-hack
     - name: Default features
-      run: cargo hack check --feature-powerset --locked --rust-version --ignore-private --workspace --all-targets
+      run: cargo hack check --feature-powerset --locked --rust-version --ignore-private --workspace --lib --bins
   minimal-versions:
     name: Minimal versions
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -64,7 +64,25 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - uses: taiki-e/install-action@cargo-hack
     - name: Default features
-      run: cargo hack check --feature-powerset --locked --rust-version --ignore-private --workspace --lib --bins
+      run: cargo hack check --feature-powerset --locked --rust-version --ignore-private --workspace --all-targets
```
Muscraft, line 67: Going from `--lib --bins` to `--all-targets` causes this job to fail as we have a `dev-dependency` on `divan` which appears to have an MSRV of ~`1.70.0` while our MSRV is `1.65.0`.

.github/workflows/ci.yml
```diff
@@ -138,3 +156,22 @@ jobs:
...
+      uses: actions/checkout@v4
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@stable
+      with:
+        toolchain: stable
+    - uses: Swatinem/rust-cache@v2
+    - name: Install cargo-tarpaulin
+      run: cargo install cargo-tarpaulin
+    - name: Gather coverage
+      run: cargo tarpaulin --output-dir coverage --out lcov
```
Muscraft, line 173: It looks like this job will fail sometimes as it can hit the default timeout (60s). Adding `--timeout 120` might solve this problem.

PR #152, opened 2024-10-04
feat: make Level::span generic over RangeBounds

SamWilsn:


src/renderer/display_list.rs
```diff
@@ -1187,16 +1187,24 @@ fn fold_prefix_suffix(mut snippet: snippet::Snippet<'_>) -> snippet::Snippet<'_>
...
-            let range_start = ann.range.start - new_start;
-            let range_end = ann.range.end - new_start;
-            ann.range = range_start..range_end;
+            let range_start = match ann.range.0 {
+                Bound::Unbounded => Bound::Unbounded,
+                Bound::Excluded(e) => Bound::Excluded(e - new_start),
+                Bound::Included(e) => Bound::Included(e - new_start),
+            };
+            let range_end = match ann.range.1 {
+                Bound::Unbounded => Bound::Unbounded,
```
SamWilsn, line 1196: Hm, I'm wondering if this needs to be `Bound::Exclusive(ann.exclusive_end(foo) - new_start)` instead.

Muscraft:
Do you mind if I ask what the reasoning behind this change was?

SamWilsn:
> Do you mind if I ask what the reasoning behind this change was?

Not at all! I tried to use a `0..=3` range, and noticed `span` wasn't generic over the types of ranges. This pull request improves the ergonomics of the `span` function so that it more closely matches the [API of `str`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=44078a26f4b7551ff6d76a7ecf32d7d0).

epage:
Could you expand on what your use case is? We expect this to be used programmatically which would generally be done with one range type and the current is most flexible.

SamWilsn:
There is no use case. This is just an ergonomic upgrade. It allows the following ranges that weren't allowed before:

```rust
Level::Error.span(2..4);    // Range (what you support now)
Level::Error.span(2..=4);   // RangeInclusive
Level::Error.span(..4);     // RangeTo
Level::Error.span(..=4);    // RangeToInclusive
Level::Error.span(2..);     // RangeFrom
Level::Error.span(..);      // RangeFull
```

If this adds too much complexity, feel free to close. I don't mind.

epage:
What i want to understand from the use case is how you are creating the spans. Are you hand writing them? If so, why?

SamWilsn:
I have a lint that always starts from the beginning of a line, which is probably the only decent example I have.

Muscraft:
Without a more specific use case, I am not sure if the added complexity is worth it at this time. I think this is something to look into more as the library progresses. 

That said, I am going to close this for now, but I greatly appreciate the PR.

Issue #108, opened 2024-03-21
Special way to underline with empty spans


epage:
The `toml` crate occasionally returns empty spans and #107 fixed it so we could render something but it might be useful to try to highlight the space between characters, rather than a character.

chengr4:
Hi @epage, 

I’m working on addressing this ticket and would appreciate your guidance.

## Background

According to your fix in #107, you added additional col annotation span allowing symbols of `Error` (`^`), `Warning` (`-`) etc., to be displayed.

---

My understanding of the goal for this ticket is illustrated by the following example:

```sh
asdf_

as_f
       'd" belongs here        
```

> '_' represents "highlight" rather than any specific symbol.

If this is the case, we would need to find a way to place the "highlight" either directly after or between characters, which is very different (I think) from the current implementation. Do you have any ideas on how we might achieve this? Or is my understanding for ticket incorrect?

epage:
To recap, `toml` sometimes highlights empty ranges (e.g. bytes at `0..0`), like at EOF or EOL when a terminating character is needed.

>  Do you have any ideas on how we might achieve this? Or is my understanding for ticket incorrect?

This is a "needs design" type of issue; no decision is made about what this should look like.

One idea I've toyed with for EOL is for us to render `\n` or `\r\n` as Dimmed and highlight those.  We could pick a character to similarly represent EOF.  In `snapbox` diffs, I chose ∅.

That is only one idea and that doesn't cover other cases of empty spans.

@estebank any thoughts?

estebank:
One option is to do [what we do in `rustc` now](https://github.com/rust-lang/rust/pull/127528): replace the ASCII control characters with their Unicode graphical representation (like `␍`). For zero-width spans in `rustc` in general, we print a one-width underline starting at the right position and "bleeding" to the right. If you [look at a representative test](https://github.com/rust-lang/rust/pull/127528/files#diff-6c752845a77286d5dc48c814eb4907b3c92498e59dc4222d537d04524351c9d2), you can see that showing *something* seems to work better than leaving the "hidden" char hidden. I had considered a long time ago using `/\` as an underline pointing in between to char positions, but ended up never doing it because there was no way of doing the same for the secondary spans (`-`).

chengr4:
If I understand correctly, now
For EOL, we have two options

```bash
# render \n or \r\n
This is a line of "code␍␊
                       ^
# OR

# print a one-width underline
This is a line of "code_
                       ^
```

[...]

PR #154, opened 2024-10-13
Fix origin position computation

KarelPeeters:
This PR fixes an issue in the computation of the position in the origin file. When a highlighted section starts at the beginning of the line, the origin position used to point to the end of the previous line, instead of the start of the current line. This resulted in outputs like this:

```
error: title
 --> origin.txt:2:4
  |
1 | aaa
2 | bbb
3 | ccc
  | ^^^ annotation
4 | ddd
  |
```

Here `origin.txt:2:4` is wrong, it should be `origin.txt:3:1`, which actually matches the highlighted setting.

As a drive-by fix, 21645ad032dd9117d811db57a5740b6a6406f94a fixes `EndLine`, which was used incorrectly in multiple places that happened to cancel out.

### DIFF
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -1149,7 +1149,7 @@ fn format_header<'a>(
                 ..
             } = item
             {
-                if main_range >= range.0 && main_range <= range.1 + end_line.len() {
+                if main_range >= range.0 && main_range < range.1 + max(*end_line as usize, 1) {
                     let char_column = text[0..(main_range - range.0).min(text.len())]
                         .chars()
                         .count();
--- a/tests/formatter.rs
+++ b/tests/formatter.rs
@@ -905,3 +905,53 @@ error: unused optional dependency
     let renderer = Renderer::plain();
     assert_data_eq!(renderer.render(input).to_string(), expected);
 }
+
+#[test]
+fn origin_correct_start_line() {
+    let source = "aaa\nbbb\nccc\nddd\n";
+    let input = Level::Error.title("title").snippet(
+        Snippet::source(source)
+            .origin("origin.txt")
+            .fold(false)
+            .annotation(Level::Error.span(8..8 + 3).label("annotation")),
+    );
+
+    let expected = str![[r#"
+error: title
+ --> origin.txt:2:4
+  |
+1 | aaa
+2 | bbb
+3 | ccc
+  | ^^^ annotation
+4 | ddd
+  |
+"#]];
+    let renderer = Renderer::plain();
+    assert_data_eq!(renderer.render(input).to_string(), expected);
+}
+
+#[test]
+fn origin_correct_mid_line() {
+    let source = "aaa\nbbb\nccc\nddd\n";
+    let input = Level::Error.title("title").snippet(
+        Snippet::source(source)
+            .origin("origin.txt")
+            .fold(false)
+            .annotation(Level::Error.span(8 + 1..8 + 3).label("annotation")),
+    );
+
+    let expected = str![[r#"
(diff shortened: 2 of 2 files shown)


KarelPeeters:
Ah, I missed those failing test cases, since they're disabled on windows:
https://github.com/rust-lang/annotate-snippets-rs/blob/c84e388949f0e06f4529de060c195e12e6531fbd/tests/fixtures/main.rs#L9-L14
Is there any reason for that, maybe CRLF issues? They seem to behave as expected for me when I just remove that `#[cfg(not(windows))]` line!

The failing test are about cases that try to point one byte past the end of the line. Previously that was indeed possible, but only in very limited situations, eg. the highlighting only worked on EOF cases anyway. See https://github.com/rust-lang/annotate-snippets-rs/issues/108#issuecomment-2290855597 for vaguely relevant discussion about this.

I'm not sure what the best way to proceed here is. Is the "one past the end of the line" use case already adequately supported by just allowing users to point at the newline characters themselves? Ideally there would be some out-of-band way to signal this case, since only using byte indices for this is tricky.

src/renderer/display_list.rs
```diff
@@ -940,9 +940,20 @@ impl<'a> CursorLines<'a> {
...
+
+impl EndLine {
+    /// The number of characters this line ending occupies in bytes.
+    pub(crate) fn len(self) -> usize {
+        match self {
+            EndLine::Eof => 0,
+            EndLine::Lf => 1,
+            EndLine::Crlf => 2,
+        }
+    }
```
Muscraft, line 956: This should probably be its own PR so it can be evaluated separately

tests/formatter.rs
```diff
@@ -905,3 +905,53 @@ error: unused optional dependency
...
+1 | aaa
+2 | bbb
+3 | ccc
+  |  ^^ annotation
+4 | ddd
+  |
+"#]];
+    let renderer = Renderer::plain();
+    assert_data_eq!(renderer.render(input).to_string(), expected);
+}
```
Muscraft, line 957: When adding new tests, they should always pass and show the current behavior. Can you please update these to show the current behavior in this commit?

src/renderer/display_list.rs
```diff
@@ -1149,7 +1149,7 @@ fn format_header<'a>(
                 ..
             } = item
             {
-                if main_range >= range.0 && main_range <= range.1 + end_line.len() {
+                if (range.0..range.1 + end_line.len()).contains(&main_range) {
```
Muscraft, line 1152: This change makes it so highlighting EOF shows the incorrect line-column. I believe the solution to this is to modify the original code to check that `main_range` is less than `range.1 + end_line` not less than or equal, as well as make it so `EOF` adds 1.
```suggestion
                if main_range >= range.0 && main_range < range.1 + max(*end_line as usize, 1) {
```

src/renderer/display_list.rs
KarelPeeters: Done, split out as #155.

tests/formatter.rs
KarelPeeters: This is surprising to me, I'm more used to "add failing test, then fix the test in the next commit". Either way is reasonable, I'll change this.

src/renderer/display_list.rs
KarelPeeters: That seems to work, thanks for the suggestion!

KarelPeeters:
I think I've addressed all of the comments, and I have rebased this PR on the current master. Let me know if I can do anything else to improve this PR!

Muscraft (approved):

Issue #97, opened 2024-03-12
Allow adding non-`Level` context to `Annotation`s


epage:
Currently, an `Annotation` is created with a `Level`.  We emulate rustc's "context" style by using a `Level::Info` but this isn't accurate (and adds a prefix).

epage:
I believe this was resolved with #195

PR #159, opened 2024-11-20
feat: Add `Level::None` for no title prefix

ZedThree:
A part solution to #97, although only really useful for the title. For the inline annotations, this currently gives no underlining.


epage:
I'm not seeing how this is related to #97.  That is about us using `Level` for Annotations when we shouldn't.  This PR is about making `Level` optional for titles.

I suspect you have a different need and that we should first start by exploring that in an issue before we decide what solution is appropriate and move on to a PR.

epage:
With v0.12, we now have [`Level::no_name`](https://docs.rs/annotate-snippets/latest/annotate_snippets/struct.Level.html#method.no_name) which I believe resolves this concern

PR #170, opened 2025-01-08
fix: Remove false positive line trimming

BurntSushi:
Previously, it was possible for a `...` to be inserted when no trimming
was actually done. For example:

```
  |
1 |   version = "0.1.0"
2 |   # Ensure that the spans from toml handle utf-8 correctly
3 |   authors = [
  |  ___________^
4 | |     { name = "Z...ALGO", email = 1 }
5 | | ]
  | |_^ RUF200
  |
```

After this fix, the `...` is no longer inserted:

```
  |
1 |   version = "0.1.0"
2 |   # Ensure that the spans from toml handle utf-8 correctly
3 |   authors = [
  |  ___________^
4 | |     { name = "ZALGO", email = 1 }
5 | | ]
  | |_^ RUF200
  |
```

This is another low confidence fix where I'm not sure that it's right.
But the implementation was previously seeming to conflate the line
length (in bytes) versus the actual rendered length. So instead of
trying to do some math to figure out whether an ellipsis should be
inserted, we just keep track of whether the code line we write was
truncated or not.

### DIFF
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -331,28 +331,28 @@ impl DisplaySet<'_> {
 
                     // On long lines, we strip the source line, accounting for unicode.
                     let mut taken = 0;
-                    let code: String = text
-                        .chars()
-                        .skip(left)
-                        .take_while(|ch| {
-                            // Make sure that the trimming on the right will fall within the terminal width.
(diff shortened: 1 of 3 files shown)


BurntSushi:
Hah, the ZALGO text doesn't seem to really render? The regression test in the PR should clarify things a bit.

PR #171, opened 2025-01-08
feat: Add `Level::None` for omitting the header

### LINKED ISSUE #167: enable a way to elide the message "header"
I'm somewhat new to this crate, so apologies for any category or vernacular errors that I make.

I'm looking at migrating [ruff](https://github.com/astral-sh/ruff) from `annotate-snippets 0.9` to `annotate-snippets 0.11`. But I'm hitting an issue where we would like to use our own headers for snippets. This worked in `0.9`, but I can't find a way to replicate it in `0.11`. To make this concrete, consider this program using `0.9`:

```rust
use annotate_snippets::{
    display_list::{DisplayList, FormatOptions},
    snippet::{Annotation, AnnotationType, Slice, Snippet, SourceAnnotation},
};

fn main() {
    let snippet = Snippet {
        title: None,
        footer: vec![
            Annotation {
                id: None,
                label: Some(
                    "Replace with `None`; initialize within function",
                ),
                annotation_type: AnnotationType::Help,
            },
        ],
        slices: vec![
            Slice {
                source: "# Docstring followed by a newline\n\ndef foobar(foor, bar={}):    \n    \"\"\"\n    \"\"\"\n",
                line_start: 1,
                origin: None,
                annotations: vec![
                    SourceAnnotation {
                        range: (
                            56,
                            58,
                        ),
                        label: "B006",
                    },
                ],
                fold: false,
            },
        ],
        opt: FormatOptions {
            color: false,
            anonymized_line_numbers: false,
            margin: None,
        },
    };
    println!("{message}", message = DisplayList::from(snippet));
}
```

It has this output:

```
  |
1 | # Docstring followed by a newline
2 |
3 | def foobar(foor, bar={}):
  |                      ^^ B006
4 |     """
5 |     """
  |
  = help: Replace with `None`; initialize within function
```

In ruff, this gets transformed slightly to include a header:

```
B006_1.py:3:22: B006 [*] Do not use mutable data structures for argument defaults
  |
1 | # Docstring followed by a newline
2 |
3 | def foobar(foor, bar={}):
  |                      ^^ B006
4 |     """
5 |     """
  |
  = help: Replace with `None`; initialize within function
```

Now consider this program, using `0.11`, which tries to mimic the above program:

```rust

fn main() {
    let source = "# Docstring followed by a newline\n\ndef foobar(foor, bar={}):    \n    \"\"\"\n    \"\"\"\n";
    let src_annotation = Level::Error.span(56..58).label("B006");
    let snippet = Snippet::source(source)
        .line_start(1)
        .annotation(src_annotation)
        .fold(false);
    let footer =
        Level::Help.title("Replace with `None`; initialize within function");
    let message =
        Level::Error.title("<TITLE>").snippet(snippet).footer(footer);
    println!("{}", Renderer::plain().render(message));
}
```

It has this output:

```
error: <TITLE>
  |
1 | # Docstring followed by a newline
2 |
3 | def foobar(foor, bar={}):
  |                      ^^ B006
4 |     """
5 |     """
  |
  = help: Replace with `None`; initialize within function
```

Which is... very close to what we had before. But it now has an `error: <TITLE>` line that I don't think can be removed. At least, it seems to correspond to a requirement to create a message by providing a `Level` with a title.

Is this an intended change to the library that isn't meant to be configurable? If so, I think our options are to either go our own way or to explore changing the formatting of our own diagnostics (which we could do, but I definitely want to see if I can de-couple upgrading `annotate-snippets` from "let's reconsider our diagnostic formatting").

I do see some other issues and PRs _seemingly_ related to this issue, but I wasn't 100% sure if this was a strict duplicate or not. So I wanted to put my use case in writing and see where that takes us.

BurntSushi:
When combined with an empty `title`, this results in the header being
omitted entirely.

This isn't what @epage suggested in #167 since changing our rendering to
pass in a custom header is a bigger refactor than what I've been able to
get to. The goal of this change was to be very small without requiring
bigger code changes on our end. Unfortunately, this is a breaking
change, so I'm not sure what the appetite is for this.

Fixes #167

### DIFF
--- a/src/renderer/display_list.rs
+++ b/src/renderer/display_list.rs
@@ -909,6 +909,7 @@ pub(crate) enum DisplayAnnotationType {
 impl From<snippet::Level> for DisplayAnnotationType {
     fn from(at: snippet::Level) -> Self {
         match at {
+            snippet::Level::None => DisplayAnnotationType::None,
             snippet::Level::Error => DisplayAnnotationType::Error,
             snippet::Level::Warning => DisplayAnnotationType::Warning,
             snippet::Level::Info => DisplayAnnotationType::Info,
--- a/src/snippet.rs
+++ b/src/snippet.rs
@@ -126,6 +126,8 @@ impl<'a> Annotation<'a> {
 /// Types of annotations.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Level {
+    /// Do not attach any annotation.
+    None,
     /// Error annotations are displayed using red color and "^" character.
     Error,
     /// Warning annotations are displayed using blue color and "-" character.
--- a/tests/formatter.rs
+++ b/tests/formatter.rs
@@ -121,6 +121,35 @@ fn test_format_title() {
     assert_data_eq!(renderer.render(input).to_string(), expected);
 }
 
+/// Tests that we can format a message *without* a header.
+///
+/// This uses `Level::None`, which is somewhat of a hacky API addition I made
+/// to our vendored copy of `annotate-snippets` in order to do exactly what
+/// this test asserts: skip the header.
+#[test]
+fn test_format_skip_title() {
+    let source =
+        "# Docstring followed by a newline\n\ndef foobar(foot, bar={}):\n    \"\"\"\n    \"\"\"\n";
+    let src_annotation = Level::Error.span(56..58).label("B006");
+    let snippet = Snippet::source(source)
+        .line_start(1)
+        .annotation(src_annotation)
+        .fold(false);
+    let message = Level::None.title("").snippet(snippet);
+
+    let expected = str![[r#"
+  |
+1 | # Docstring followed by a newline
+2 |
+3 | def foobar(foot, bar={}):
+  |                      ^^ B006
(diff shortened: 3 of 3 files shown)


ZedThree:
I have a practically identical change in #159 for very similar reasons :)

BurntSushi:
Ah nice, whoops, missed that!

ndmitchell:
I'm also after pretty much the same change. Any chance this could be reviewed?

Muscraft:
I am sorry for not reviewing or commenting on this before now. `annotate-snippets` is currently in the middle of a major rewrite, as part of the [project goal](https://rust-lang.github.io/rust-project-goals/2025h1/annotate-snippets.html) to use it for `rustc`'s output. This has been my primary focus over the past few months. That said, I plan to add support for this when the new version is released (hopefully in the next few weeks).

epage:
With v0.12, we now have [`Level::no_name`](https://docs.rs/annotate-snippets/latest/annotate_snippets/struct.Level.html#method.no_name) which I believe resolves this concern

PR #172, opened 2025-01-09
feat: Add `Renderer::cut_indicator`

BurntSushi:
This adds a new API for overriding the use of `...` to indicate a cut or
trimmed line. In the case of Ruff, we didn't want to use `...` since
`...` is valid Python code. It could be rather confusing in some cases
where `...` would be ambiguous between "line was cut here" and "this is
what the actual line read as." I think this can happen with _any_
indicator of course, but for Python specifically, it's pretty likely to
happen with `...`.

The new API here is somewhat sub-optimal in that it requires a
`&'static str`. I did this because of the constraints imposed by
a `Renderer`'s `const` constructor.


epage:
It might be good for there to be an issue for exploring this before moving onto a PR as this might run counter to some other goals
- Converging on a common set of best practices / being opinionted
- Handling unicode vs ascii via a single flag rather than attempting to make the whole layout engine themable

BurntSushi:
Aye, fair enough! I opened #174 for that.

Issue #184, opened 2025-03-17
Dividing display_list.rs file.


PatelTirth25:
Currently, display_list.rs file is of 1700+ lines, which might be hard to read. So I guess it would be a good idea to divide file into folder as it can improve readability, easier collaborations and many more.

epage:
#191 proposed splitting things into a lot of very small files in an even deeper nested mod.

An analogy for code organization that I've adopted is icache.  CPUs are able to execute faster based on instructions executed together being close together.  I think we do similar when reasoning about code.  The more tightly coupled two pieces of code are, the closer they should be to each other.  With how small some of the `mod`s are, I'm worried about how well this is being met.

One way to consider this is how much private (non-`pub(*)`) code can exist within a `mod`.  If there is a lot, then that code has no coupling to the outside.  If its all `pub(crate)` (or `pub(super`, `pub(in ...)`, etc) then its a very shallow API and likely has tight coupling and would be better served being next to what is using it.

We likely should wait for any refactorings until after we've finished this sync with rustc and been able to switch rustc over to annontate-snippets.  This is to avoid efforts in these code paths from being conflict heavy and the current code is meant to be very similar to the rustc code for logic to be moved back and forth.

I also suspect this is something that should be done more organically, based on experience and need.

PR #191, opened 2025-03-17
refactor: Display

PatelTirth25:
This PR refactors the display_list.rs file, which was over 1700 lines, into a more manageable structure by splitting it into separate modules. (https://github.com/rust-lang/annotate-snippets-rs/issues/184)

Folder structure:

src/renderer/display
├── constants.rs
├── cursor_line.rs
├── display_annotations.rs
├── display_header.rs
├── display_line.rs
├── display_list.rs
├── display_mark.rs
├── display_set.rs
├── display_text.rs
├── end_line.rs
└── mod.rs


epage:
I lean towards closing this
- There are a lot of conflicts and likely to be more
- If we do a split like this, we should probably discuss how things should be split in #184

Issue #197, opened 2025-04-16
`Origin::label` doesn't make sense


epage:
```rust
                .element(
                    Origin::new("$SRC_DIR/core/src/cmp.rs")
                        .line(334)
                        .char_column(14)
                        .primary(true)
                        .label("...because it uses `Self` as a type parameter")

                )
```
seems to create
```
  --> $SRC_DIR/core/src/cmp.rs:334:14
   |
   = note: ...because it uses `Self` as a type parameter
```

Issue #196, opened 2025-04-16
`Level.header` vs `Level.title` is confusing


epage:
As part of this, we should keep in mind #118

PR #195, opened 2025-04-16
New api

Muscraft:
This PR moves `annotate-snippets` internals to better match `rustc`,  better matches `rustc`'s output, adds suggestions, and moves to a new API. The new API is subject to change as user feedback comes in on how it can be improved.

PR #198, opened 2025-04-16
fix!: Iterate on the API

### LINKED ISSUE #175: `Renderer::line_no` is a confusing name
What is `no`, `nomber`?

Is `line_num` / `line_number` the best name?  I feel like this is the `gutter` style but its also used for `...`.  Unsure what to call that.

We should probably change this to something else.  If we want, we can smooth out the upgrade path by offering the name in a current version, deprecating the old name, and then removing the old name in a new version.

epage:
This is basically review feedback on top of #195

Some things were deferred out, like #196, #197

Fixes #175

BREAKING CHANGE: Renamed `Renderer::line_no` to `Renderer::line_num`


Muscraft (approved):
These changes look good to me!

PR #216, opened 2025-06-12
fix: Match how rustc handles annotating newlines

Muscraft:
Note: This is another change related to matching `rustc`'s output


epage:
I thought we were a bit unsure of this one?

PR #217, opened 2025-06-12
fix: Remove unneeded annotation sorting

Muscraft:
Note: This is another change related to matching rustc's output


epage:
Should we note this in the documentation?

PR #218, opened 2025-06-24
fix!: Remove of Origin::label

### LINKED ISSUE #197: `Origin::label` doesn't make sense

Muscraft:
resolves #197


epage (approved):
Feel free to merge when its ready

Issue #208, opened 2025-06-05
`Element::Origin` and `Snippet::origin` are unclear when to use either


epage:
This is based on what is currently in master

epage:
I believe this was resolved in #227

PR #221, opened 2025-06-24
refactor: Use Origin for Snippet::origin

Muscraft:
This partially addresses #208 by making `Snippet::origin` use `Origin` internally. This makes the distinction a little clearer: you should use `Snippet::origin` if you are using a `Snippet` and `Origin` on its own if you need the file details.

### DIFF
--- a/src/snippet.rs
+++ b/src/snippet.rs
@@ -164,7 +164,7 @@ pub struct Title<'a> {
 /// A source view [`Element`] in a [`Group`]
 #[derive(Clone, Debug)]
 pub struct Snippet<'a, T> {
-    pub(crate) origin: Option<&'a str>,
+    pub(crate) origin: Option<Origin<'a>>,
     pub(crate) line_start: usize,
     pub(crate) source: &'a str,
     pub(crate) markers: Vec<T>,
@@ -200,6 +200,11 @@ impl<'a, T: Clone> Snippet<'a, T> {
 
     /// The location of the [`source`][Self::source] (e.g. a path)
     ///
+    /// If only a location is provided (i.e. a `String`) then the rest of the
+    /// [`Origin`] is inferred (e.g. line and column numbers).  To adjust line
+    /// numbers, consider using [`Snippet::line_start`] instead as it will also
+    /// adjust line numbers for the [`Snippet::source`].
+    ///
     /// <div class="warning">
     ///
     /// Text passed to this function is considered "untrusted input", as such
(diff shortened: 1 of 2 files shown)


src/snippet.rs
```diff
@@ -207,8 +207,8 @@ impl<'a, T: Clone> Snippet<'a, T> {
     /// not allowed to be passed to this function.
     ///
     /// </div>
-    pub fn origin(mut self, origin: &'a str) -> Self {
-        self.origin = Some(origin);
+    pub fn origin(mut self, origin: impl Into<Origin<'a>>) -> Self {
+        self.origin = Some(origin.into());
         self
```
epage, line 212: Not too sure how this makes it clearer.

Also, wht does uit mean to specify line, char_column, or primary on an Origin in a Snippet?

Muscraft: `is_primary` is respected unless the `Origin` is the "primary origin" for a `Group` (I think).

`line` is respected if it is set.

`char_column` is respected if `line` is set; if `line` is not set, it is overridden.

src/snippet.rs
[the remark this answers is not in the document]
Muscraft: Let me know what you think of the docs I added

src/snippet.rs
```diff
@@ -375,8 +379,16 @@ impl<'a> Patch<'a> {
     }
 }
 
-/// The location of the [`Snippet`] (e.g. a path)
-#[derive(Clone, Debug)]
+/// The location of the [`Snippet`] (e.g. a path).
+///
+/// This should be used if you want to display the file details by
+/// itself, or you want to set the line number and column explicitly.
+/// Otherwise, it is better to use [`Snippet::origin`].
```
epage, line 386: This should be clarified that you mean that this is a parallel item to `Snippet` (can't remember all of the terms and not digging in to provide a suggestion atm)

src/snippet.rs
```diff
@@ -375,8 +380,16 @@ impl<'a> Patch<'a> {
     }
 }
 
-/// The location of the [`Snippet`] (e.g. a path)
-#[derive(Clone, Debug)]
+/// The location of the [`Snippet`] (e.g. a path).
+///
+/// This should be used if you want to set the line number and column
+/// explicitly for a [`Snippet`], or if you need to render a location without
+/// an accompanying [`Snippet`].
```
epage, line 387:
```suggestion
/// The referenced location (e.g. a path)
///
/// If you have source available, see instead [`Snippet`]
```

src/snippet.rs
```diff
@@ -164,7 +164,7 @@ pub struct Title<'a> {
 /// A source view [`Element`] in a [`Group`]
 #[derive(Clone, Debug)]
 pub struct Snippet<'a, T> {
```
epage, line 166:
```suggestion
/// A source view [`Element`] in a [`Group`]
///
/// If you do not have [source][Snippet::source] available, see instead [`Origin`]
#[derive(Clone, Debug)]
pub struct Snippet<'a, T> {
```

Muscraft:
Closing in favor of #227, as it does a better job of addressing #208.

PR #230, opened 2025-06-27
Add suppost for Hyperlinks on IDs

Muscraft:
`annotate-snippets` needs to support hyperlinks on IDs as `rustc` has an unstable feature for them, and it will more than likely be stabilized in the future.


Cargo.toml
```diff
@@ -132,6 +132,7 @@ unicode-width = "0.2.0"
 [dev-dependencies]
 annotate-snippets = { path = ".", features = ["testing-colors"] }
 anstream = "0.6.13"
+anstyle-svg = "0.1.8"
```
epage, line 135: why the direct dep?

PR #232, opened 2025-06-27
fix: Take Into<Cow> instead of &str

### LINKED ISSUE #200: Owned variants of APIs where &str is used
I've been using this crate for personal project of mine, and I've sometimes found myself wanting to do something like this this:

```rust
fn make_message(token: &Token) -> Message<'_> {
    Level::Error.header(format!("unhandled token `{}`", token.src))
        .[...]
}
```

For obvious reasons, this doesn't directly work. To make it work the error string has to be stored out of line elsewhere, but this isn't always convenient (at least, it isn't for my design). It would be easier if there was a variant of (at least some) of the APIs which accepts a `String` instead, and the fields internally store a `Cow<'src, 'str>` instead of a `&'src str`.

Incidentally, this also might help people who have use cases that would have wanted #101, since it would be easier to directly format into a `String` that is stored by the `Message`.

Muscraft:
Fixes #200

### DIFF
--- a/src/level.rs
+++ b/src/level.rs
@@ -2,8 +2,9 @@
 
 use crate::renderer::stylesheet::Stylesheet;
 use crate::snippet::{ERROR_TXT, HELP_TXT, INFO_TXT, NOTE_TXT, WARNING_TXT};
-use crate::Title;
+use crate::{OptionCow, Title};
 use anstyle::Style;
+use std::borrow::Cow;
 
 /// Default `error:` [`Level`]
 pub const ERROR: Level<'_> = Level {
@@ -38,7 +39,7 @@ pub const HELP: Level<'_> = Level {
 /// [`Title`] severity level
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Level<'a> {
-    pub(crate) name: Option<Option<&'a str>>,
+    pub(crate) name: Option<Option<Cow<'a, str>>>,
     pub(crate) level: LevelInner,
 }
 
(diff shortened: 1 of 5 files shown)


src/level.rs
```diff
@@ -56,9 +57,9 @@ impl<'a> Level<'a> {
     /// not allowed to be passed to this function.
     ///
     /// </div>
-    pub fn text(self, text: Option<&'a str>) -> Level<'a> {
+    pub fn text(self, text: Option<impl Into<Cow<'a, str>>>) -> Level<'a> {
```
epage, line 60: A place where `OptionCow` is unused

src/snippet.rs
```diff
@@ -407,6 +411,38 @@ impl<'a> Origin<'a> {
...
 }
 
+impl<'a> From<Cow<'a, str>> for Origin<'a> {
+    fn from(origin: Cow<'a, str>) -> Self {
+        Self::new(origin)
+    }
+}
+
+#[derive(Debug)]
+pub struct OptionCow<'a>(Option<Cow<'a, str>>);
```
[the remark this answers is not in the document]
epage: We use `Option<into Cow>` in places, like https://github.com/rust-lang/annotate-snippets-rs/pull/232/files#r2172869778

tests/formatter.rs
```diff
@@ -2448,7 +2448,7 @@ fn secondary_title_no_level_text() {
         )
         .element(
             Level::NOTE
-                .text(None)
+                .text(None::<&str>)
```
epage, line 2451: Is this still needed?

Muscraft: sadly yes

epage: How come and then why does `OptionCow` exist?  I thought it was to bypass this problem

Muscraft: `::<&str>` is needed because the type of `None` can't be inferred, as the `From` impl for `Option` is generic over `T: Into<Cow<'a, str>>`. This problem would still exist if we were to write out the different `From`. 

`OptionCow` exists to solve the more common case of passing in a literal without also needing to write `Some()`.

epage: I have a way of solving this with `None` with a new type but I thought I had a way of solving this without `None` without any new types but my ideas aren't working and not finding any breadcrumbs (in clap I went from `impl Into<Option<&str>>` straight to the solution for `None`).

PR #235, opened 2025-06-30
feat: Allow setting the primary level for a group

Muscraft:
This should allow users to create a `Group` without a `Title` and control the primary `Level` for that `Group`.

### DIFF
--- a/examples/elide_header.svg
+++ b/examples/elide_header.svg
@@ -3,7 +3,7 @@
     .fg { fill: #AAAAAA }
     .bg { background: #000000 }
     .fg-bright-blue { fill: #5555FF }
-    .fg-bright-cyan { fill: #55FFFF }
+    .fg-bright-red { fill: #FF5555 }
     .container {
       padding: 0 10px;
       line-height: 18px;
@@ -27,7 +27,7 @@
 </tspan>
     <tspan x="10px" y="82px"><tspan class="fg-bright-blue bold">3</tspan><tspan> </tspan><tspan class="fg-bright-blue bold">|</tspan><tspan> def foobar(door, bar={}):</tspan>
 </tspan>
-    <tspan x="10px" y="100px"><tspan>  </tspan><tspan class="fg-bright-blue bold">|</tspan><tspan>                      </tspan><tspan class="fg-bright-cyan bold">^^</tspan><tspan> </tspan><tspan class="fg-bright-cyan bold">B006</tspan>
+    <tspan x="10px" y="100px"><tspan>  </tspan><tspan class="fg-bright-blue bold">|</tspan><tspan>                      </tspan><tspan class="fg-bright-red bold">^^</tspan><tspan> </tspan><tspan class="fg-bright-red bold">B006</tspan>
 </tspan>
     <tspan x="10px" y="118px"><tspan class="fg-bright-blue bold">4</tspan><tspan> </tspan><tspan class="fg-bright-blue bold">|</tspan><tspan>     """</tspan>
 </tspan>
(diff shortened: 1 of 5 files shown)


examples/elide_header.svg
epage, line 1: The documentation says the fallback is "error" but a cyan shouldn't be an error.
https://github.com/rust-lang/annotate-snippets-rs/blob/c195b836696312ee80880c1fe4edf6003222aade/src/snippet.rs#L265-L266

src/snippet.rs
```diff
@@ -44,6 +48,14 @@ impl<'a> Group<'a> {
         self
     }
 
+    /// Set the primary [`Level`] for this [`Group`]. If not specified, the
+    /// [`Level`] of the first [`Title`] in the [`Group`] will be used, if no
+    /// [`Title`] is present, it will default to [`Level::ERROR`].
+    pub fn primary_level(mut self, level: Level<'a>) -> Self {
```
epage, line 54: The following documentation should be updated
https://github.com/rust-lang/annotate-snippets-rs/blob/c195b836696312ee80880c1fe4edf6003222aade/src/snippet.rs#L265-L266

src/snippet.rs
```diff
@@ -44,6 +48,14 @@ impl<'a> Group<'a> {
         self
     }
 
+    /// Set the primary [`Level`] for this [`Group`]. If not specified, the
+    /// [`Level`] of the first [`Title`] in the [`Group`] will be used, if no
+    /// [`Title`] is present, it will default to [`Level::ERROR`].
```
epage, line 53: This needs a blank line after the first sentence so rustdoc can use that for the summary, rather than the whole block of text

src/snippet.rs
epage, line 54: a bit hesitant with the name but we can worry about that in a more holistic review of the API rather than the pigeonhole of this specific PR

examples/elide_header.svg
Muscraft: This comment was correct in that the `Level` of the first `Title` in a `Group` would be used, not the first `Element` if it is a `Title`. I added a comment that changes this to act how you read it and updated the docs accordingly.

src/snippet.rs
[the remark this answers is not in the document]
Muscraft: This comment was correct in that the `Level` of the first `Title` in a `Group` would be used, not the first `Element` if it is a `Title`. I added a commit that changes this to act as you read it and updated the docs accordingly.

examples/elide_header.svg
epage: So because there was a help further down, that became the level?

Muscraft: Yep!

src/snippet.rs
```diff
@@ -262,10 +276,21 @@ impl<'a> Annotation<'a> {
...
 pub enum AnnotationKind {
-    /// Color to the [`Level`] the first [`Title`] in [`Group`]. If no [`Title`]
-    /// is present, it will default to `error`.
+    /// An [`Annotation`] which will use the "primary" underline character ('^'
+    /// for [`OutputTheme::Ascii`], and '━' for [`OutputTheme::Unicode`]) and
+    /// be colored according to the primary [`Level`] of the [`Group`], see
+    /// [`Group::primary_level`] for details about how this is set.
+    ///
+    /// [`OutputTheme::Ascii`]: crate::renderer::OutputTheme
+    /// [`OutputTheme::Unicode`]: crate::renderer::OutputTheme
```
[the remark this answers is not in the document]
epage: Another way to refer to this is this is the root cause of the group.

PR #243, opened 2025-07-02
fix: Split out `Message` from `Title`

### LINKED ISSUE #118: Rename `title` to `header`
This is inspired by #114 where someone wanted to include a lot of details in the "title" but assumed they couldn't.

epage:
In #231, `Level::header` was split into `Level::title` and `Level::pre_styled_title` but
- When to use each wasn't clear
- "title" suggests a stronger role within the output than it is which can be confusing
- `Title` contains an `id` even when it will go unused

This reworks things into
- `Level::title() -> Title`: used for the first message within a group
  - uses a built-in styling
  - can have an `id`
- `Level::message() -> Message`: used for additional text to be shown to the user
  - callers can provide their own styling
  - cannot have an `id`

Fixes #118

### DIFF
--- a/src/level.rs
+++ b/src/level.rs
@@ -2,7 +2,7 @@
 
 use crate::renderer::stylesheet::Stylesheet;
 use crate::snippet::{ERROR_TXT, HELP_TXT, INFO_TXT, NOTE_TXT, WARNING_TXT};
-use crate::{OptionCow, Title};
+use crate::{Message, OptionCow, Title};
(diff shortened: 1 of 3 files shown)


epage:
@Muscraft not too sure what I'm missing as to why the test failure is happening.  I left `Message` out of some places or assumed some values but, looking at how its supposed to be used, these seem correct.

Muscraft (approved):

PR #251, opened 2025-07-03
fix: Show Group/File start for Snippets without a path

### LINKED ISSUE #244: Snippets without a path don't have a proper group start (unicode)
```rust
#!/usr/bin/env nargo
---
[dependencies]
annotate-snippets.git = "https://github.com/rust-lang/annotate-snippets-rs.git"
---

use annotate_snippets::{renderer::OutputTheme, AnnotationKind, Group, Level, Renderer, Snippet};

fn main() {
    // Taken from: https://docs.python.org/3/library/typing.html#annotating-callable-objects
    let source = "def __call__(self, *vals: bytes, maxlen: int | None = None) -> list[bytes]: ...";
    let group = Group::with_title(Level::ERROR.title("")).element(
        Snippet::source(source).annotation(AnnotationKind::Primary.span(4..12).label("annotation")),
    );

    println!(
        "{}",
        Renderer::styled()
            .theme(OutputTheme::Unicode)
            .term_width(83)
            .render(&[group])
    );
}

```

Image

Muscraft:
fixes #244


tests/formatter.rs
```diff
@@ -2803,6 +2803,7 @@ error:
 1 | def __call__(self, *vals: bytes, maxlen: int | None = None) -> list[bytes]: ...
   |     ^^^^^^^^ annotation
   |
+ :::
```
epage, line 2806: ditto

tests/formatter.rs
```diff
@@ -2811,10 +2812,11 @@ error:
 
     let expected_unicode = str![[r#"
 error: 
-  │
+  ╭▸ 
 1 │ def __call__(self, *vals: bytes, maxlen: int | None = None) -> list[bytes]: ...
   │     ━━━━━━━━ annotation
   │
+  ⸬
```
epage, line 2819: ditto

tests/color/issue_9.term.svg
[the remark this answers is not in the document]
epage: Looking at #9, manual folding like this seems unrelated to the issue and so we don't seem to have any requirements blocking this change.  We may want to update the example to stop manual folding.

PR #260, opened 2025-07-31
chore: Don't allow Title in element

Muscraft:
Having `Title` as an `Element` did not make sense, as `Title` is only meant to be used with `Group::with_title`, and `Message` should be used if you want similar output. Given that within our own tests, we were using `Title` when we should've been using `Message`, having both would have no doubt led to confusion for users over what to use.

Note: this adds a bit of duplication for how we handle continuations; the next few PRs should hopefully address some of it.

### DIFF
--- a/examples/highlight_source.rs
+++ b/examples/highlight_source.rs
@@ -22,7 +22,7 @@ fn main() {}
                 ),
         )
         .element(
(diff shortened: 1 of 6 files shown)


epage:
How come you didn't do the public/private enums?

PR #264, opened 2025-08-04
fix: Draw file start continuation if not first in group

Muscraft:


src/renderer/mod.rs
```diff
@@ -331,6 +331,7 @@ impl Renderer {
                                     &annotated_lines,
                                     max_depth,
                                     peek.is_some() || (g == 0 && group_len > 1),
+                                    i == 0,
```
epage, line 334: can you create a variable for this before the `match` that gets reused?

epage (approved):
Once resolved, feel free to merge

PR #268, opened 2025-08-04
test: annotate newline empty span

### LINKED ISSUE #135: Problem with Unix line feed annotation using empty ranges
Hello, thanks for your amazing work on this crate!

I recently encountered a problem while trying to annotate Unix line feeds (`\n`). When annotating line feeds with an empty `Range`, I got the following error:
```
error: bad
--> test.txt:1:1
 |
 |
```

This looks like a bug because doing the same thing on letters or digits works as expected:
```
error: bad
 --> test.txt:1:1
  |
1 | 111
  | ^
  |
```
Also, giving a non-empty `Range` solves the problem and outputs the annotation as expected:
```
error: bad
 --> test.txt:1:1
  |
1 | ...
  | ^
  |
```

Here is the code used to reproduce the problem (which occurs with the latest release). `test.txt` is a simple text file containing only Unix line feeds (`\n`):

```rust

pub fn reproduce_bug() {
    let message = Level::Error.title("bad").snippet(
        Snippet::source(include_str!("../test.txt"))
            .origin("test.txt")
            .fold(true)
            .annotation(Level::Error.span(0..0)),
    );
    anstream::println!("{}", renderer.render(message));
}
```

Muscraft:
Issue #135 was opened over a year ago, and during that time, a change to `annotate-snippets` fixed the issue, but it was never closed. Since the issue is fixed, I figured we should add a regression test to ensure consistent behavior in the future.

Fixes #135

### DIFF
--- a/tests/formatter.rs
+++ b/tests/formatter.rs
@@ -440,6 +440,37 @@ error:
 }
 
 #[test]
+fn annotate_newline_empty_span() {
+    let message = &[Group::with_title(Level::ERROR.title("bad")).element(
+        Snippet::source("\n\n\n\n\n\n\n")
+            .path("test.txt")
+            .annotation(AnnotationKind::Primary.span(0..0)),
+    )];
+
+    let expected_ascii = str![[r#"
+error: bad
+ --> test.txt:1:1
+  |
+1 |
+  | ^
+"#]];
+
+    let renderer = Renderer::plain();
+    assert_data_eq!(renderer.render(message), expected_ascii);
+
+    let expected_unicode = str![[r#"
+error: bad
+  ╭▸ test.txt:1:1
+  │
+1 │
+  ╰╴━
+"#]];
+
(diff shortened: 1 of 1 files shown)


tests/formatter.rs
```diff
@@ -439,6 +439,37 @@ error:
...
+
+    let renderer = Renderer::plain();
+    assert_data_eq!(renderer.render(message), expected_ascii);
+
+    let expected_unicode = str![[r#"
+error: bad
+  ╭▸ test.txt:1:1
+  │
+1 │
+  ╰╴━
```
epage, line 466: Nothing gets underlined in unicode?

Muscraft: I am not sure what you mean, the [underline primary character](https://github.com/rust-lang/annotate-snippets-rs/blob/66e67f506b3c1bb98f2853b97c4f3509e38e1227/src/renderer/mod.rs#L2570) (`━`) is there.

PR #323, opened 2025-10-20
fix: Properly calculate the max line num for suggestions

Muscraft:
There were some cases where we were not calculating the correct max line number for suggests, this PR addresses those problems.


src/snippet.rs
```diff
@@ -134,7 +134,7 @@ impl From<Padding> for Element<'_> {
 }
 
 /// A whitespace [`Element`] in a [`Group`]
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
```
epage: Are we sure we want to have this from an API perspective?

PR #343, opened 2025-11-11
Fix & test for correct highlighting of strings with shared prefix

faculerena:
Applies here the fix suggested in [this PR](https://github.com/rust-lang/rust/pull/148061).

[Original issue](https://github.com/rust-lang/rust/issues/148070).

### DIFF
--- a/src/renderer/source_map.rs
+++ b/src/renderer/source_map.rs
@@ -779,6 +779,19 @@ pub(crate) fn as_substr<'a>(
     original: &'a str,
     suggestion: &'a str,
 ) -> Option<(usize, &'a str, usize)> {
+    // Case for import paths where the suggestion shares a prefix with the original.
+    // Without this, suggesting `std::sync` for `sync` would incorrectly highlight `td::s`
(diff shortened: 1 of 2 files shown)


faculerena:
I'm not sure, but I sent it to talk about that. I agree that the path part of the fix looks like it should be in `rustc`, but the issue will persist here for other strings (the prefix for origin & suggestion matches then the shared_prefix is "wrong" (not really wrong, but it does not feel right))

Muscraft:
I took some time to try and find a solution that didn't require adding knowledge of `rust` syntax and ended up finding one (#345). For good measure, I ran the fix against `rust-lang/rust`'s testsuite, as well as a test for `std::sync`, and it passed. I'll open a PR updating `annotate-snippets` in `rust-lang/rust` once #345 is merged and I have released a new version.

faculerena:
Amazing man, the other fix will probably stay in rustc stable if annotate-snippets is not stabilized before 1.93, then that solution will work. Thanks!

PR #358, opened 2025-12-24
test: Match anstyle-svg background color fix

Muscraft:
[`anstyle-svg@0.1.11`](https://github.com/rust-cli/anstyle/blob/main/crates/anstyle-svg/CHANGELOG.md#0111---2025-09-17) fixed an issue with the background color not applying by switching from `.bg { background: <color> }` to `.bg { fill: <color> }` when specifying the background color. For some reason, this change did not cause the SVG tests to fail, and only shows up when an SVG test's output gets updated. This behaviour means that any change to an SVG test's output would include this unrelated change, muddying the git history. To address this issue, I switched all SVG tests to use the new background format.

### DIFF
--- a/examples/custom_error.svg
+++ b/examples/custom_error.svg
@@ -1,7 +1,7 @@
 <svg width="751px" height="128px" xmlns="http://www.w3.org/2000/svg">
   <style>
     .fg { fill: #AAAAAA }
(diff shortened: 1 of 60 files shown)


epage:
Asserting on svgs specifically ignores head

PR #365, opened 2026-02-01
feat: Add `no_std` support

### LINKED ISSUE #359: Add `no_std` support (with `alloc`)
Would be nice to be able to use this in `no_std` environments.

From a quick look, all of the current dependencies should work with `no_std`.

There's one `HashMap` usage, though looks like it could be replaced with a `BTreeMap` without altering behavior.

CathalMullan:
Resolves #359

Only meaningful change is the swap from HashMap to BTreeMap in `render.rs`.
Running the benchmarks locally showed no real perf impact (within 1%).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
     permissions:
       contents: none
     name: CI
-    needs: [test, msrv, lockfile, docs, rustfmt, clippy, minimal-versions]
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
epage: Please add the following to `src/lib.rs`:
```rust
#![warn(clippy::std_instead_of_core)]
#![warn(clippy::std_instead_of_alloc)]
```

Issue #405, opened 2026-04-19
Invalid patch highlighting for indented code


0x2a-42:
I noticed in my [Lua linter](https://github.com/0x2a-42/yutu), that the highlighting of patches is wrong, if the code is indented. It works correctly if the code is not indented.
```lua
do
  local function f()
    print()
  end
end
```

<img width="740" height="272" alt="Image" src="https://github.com/user-attachments/assets/5155db83-ef7d-426c-ba1f-cecb26f96e05" />

Here is a minimal reproducer.
```rust
use annotate_snippets::{Group, Level, Patch, Renderer, Snippet};

fn main() {
    let source = "do\n  local function f()\n    print()\n  end\nend\n";
    let delete_span = 5..41;
    let group = Group::with_title(Level::HELP.secondary_title("delete the definition"))
        .element(Snippet::source(source).patch(Patch::new(delete_span, "")));
    let renderer = Renderer::styled();
    println!("{}", renderer.render(&[group]));
}
```

0x2a-42:
The highlighting works without indentation because here the following code is used, that removes whole lines.
https://github.com/rust-lang/annotate-snippets-rs/blob/5b99daaa4ad1964e4da27525a4209e50a0340404/src/renderer/render.rs#L1498-L1521

If a line is only partially removed the colorization happens here instead.
https://github.com/rust-lang/annotate-snippets-rs/blob/5b99daaa4ad1964e4da27525a4209e50a0340404/src/renderer/render.rs#L1762-L1768

This code assumes that there is at least one addition in the diff, which should be the case if a line was partially removed.
https://github.com/rust-lang/annotate-snippets-rs/blob/5b99daaa4ad1964e4da27525a4209e50a0340404/src/renderer/render.rs#L1684-L1710

This assumption is used in the calculation of the `row` value.
https://github.com/rust-lang/annotate-snippets-rs/blob/5b99daaa4ad1964e4da27525a4209e50a0340404/src/renderer/render.rs#L1730

When only whitespaces remain in the partially removed line, this addition is however not written to the buffer. So in my example the values of `row` are [2, 2, 3] instead of [2, 3, 4], which explains why line 3 gets the colorization that should be applied to line 4.

I see three possible solutions.
1. Treat partially removed lines where only whitespaces remain as completely removed lines.
2. Somehow show trailing whitespace additions in the diff.
3. Check if there was an addition to decide if `2` or `1` should be subtracted from `row`.

PR #407, opened 2026-04-23
Fix removal highlighting for multiple indented lines

0x2a-42:
Fix for #405.

I think this change should cause partially removed lines, where only whitespace characters remain, to be treated like completely removed lines.

All tests pass except for `multiple_highlight_duplicated`, which does however not show any visual changes. The only difference is that previously some whitespace characters where not highlighted in red.

### DIFF
--- a/src/renderer/source_map.rs
+++ b/src/renderer/source_map.rs
@@ -556,10 +556,8 @@ impl<'a> SourceMap<'a> {
                 }
             }
         }
-        // remove trailing newlines
-        while buf.ends_with('\n') {
-            buf.pop();
-        }
+        // remove trailing whitespace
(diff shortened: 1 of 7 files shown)


Muscraft:
Sorry for the delay in reviewing this PR; my day job has been very busy lately, leaving me little time for open source work.

I want to thank you for taking the time to investigate the issue! It was a big help while I was digging into the issue and your solution! While I was digging into it, I realized that switching to removing all trailing whitespace only addressed a symptom of the underlying issue, which led me to dig deeper and eventually open #413. I am going to close this PR in favor of that one. Again, thank you for all the time and effort you put into #405 and this PR!

PR #434, opened 2026-07-05
fix(highlight_source): Apply styling to multiline annotations

### LINKED ISSUE #427: `highlight(true)` has no effect on multiline annotations
With this snippet (the test `ann_multiline.rs` but with `highlight_source(true)` added):

```rs
    let source = r#"                        if let DisplayLine::Source {
                            ref mut inline_marks,
                        } = body[body_idx]
"#;

    let input = &[Level::ERROR
        .primary_title("pattern does not mention fields `lineno`, `content`")
        .id("E0027")
        .element(
            Snippet::source(source)
                .path("src/display_list.rs")
                .line_start(139)
                .fold(false)
                .annotation(
                    AnnotationKind::Primary
                        .span(31..128)
                        .label("missing fields `lineno`, `content`")
                        .highlight_source(true),
                        // ^
                ),
        )];
```

The annotated portion of the source code isn't actually highlighted:

Image

---

Tested with version `0.12.16` and `rustc 1.96.0`.

momomuchu:
Fixes #427.

highlight_source was silently ignored on multiline annotations
(MultilineStart/MultilineLine/MultilineEnd). Two things conspired to
cause this:

- render_source_line has a couple of fast paths (the "short start with
  no leading text" case, and the "only connector lines, nothing to
  show" case) that can return before the per-annotation styling logic
  ever runs.
- Even when those aren't hit, the per-annotation match arm that
  applies highlight_source is placed after the arm that draws the
  MultilineStart/MultilineEnd connector, so it never matches for those
  variants, since Rust match arms are first-match-wins.

Styling multiline annotations up front, before any of that, sidesteps
both issues instead of trying to thread a fix through three separate
spots.

Added a regression test (ascii + unicode goldens) using the repro from
the issue.

### DIFF
--- a/src/renderer/render.rs
+++ b/src/renderer/render.rs
@@ -876,6 +876,39 @@ fn render_source_line(
         return vec![];
     }
 
+    // Apply `highlight_source` styling for multiline annotations here, before the fast paths
+    // below (the "short start" case just below, and the "only connector lines" case further
+    // down) get a chance to return without ever reaching the per-annotation styling logic that
+    // handles it for single-line annotations. Even when those fast paths aren't taken, the
+    // per-annotation styling logic matches on `MultilineStart`/`MultilineEnd` first to draw the
+    // horizontal connector, which made the `highlight_source` arm structurally unreachable for
+    // multiline annotations (see issue #427: `highlight_source` was silently ignored on them).
+    for ann in &line_info.annotations {
+        if !ann.highlight_source {
+            continue;
+        }
+        let (start_char, end_char) = match ann.annotation_type {
+            LineAnnotationType::Singleline => continue,
+            // The span starts partway through this line and runs to the end of it.
+            LineAnnotationType::MultilineStart(_) => {
+                let rest_of_line = line_info.line[ann.start.byte..].chars().count();
+                (ann.start.char, ann.start.char + rest_of_line)
+            }
+            // The span runs from the start of this line up to `end`.
+            LineAnnotationType::MultilineEnd(_) => (0, ann.end.char),
+            // The span fully covers this line.
+            LineAnnotationType::MultilineLine(_) => (0, line_info.line.chars().count()),
+        };
+        let underline = renderer.decor_style.underline(ann.is_primary());
+        buffer.set_style_range(
+            line_offset,
+            (code_offset + start_char).saturating_sub(left),
+            (code_offset + end_char).saturating_sub(left),
+            underline.style,
+            ann.is_primary(),
+        );
+    }
+
     // Special case when there's only one annotation involved, it is the start of a multiline
     // span and there's no text at the beginning of the code line. Instead of doing the whole
     // graph:
--- a/tests/color/ann_multiline_highlight_source.unicode.term.svg
+++ b/tests/color/ann_multiline_highlight_source.unicode.term.svg
@@ -1,0 +1,40 @@
+<svg width="740px" height="164px" xmlns="http://www.w3.org/2000/svg">
+  <style>
+    .fg { fill: #AAAAAA }
+    .bg { fill: #000000 }
+    .fg-bright-blue { fill: #5555FF }
+    .fg-bright-red { fill: #FF5555 }
+    .container {
+      padding: 0 10px;
+      line-height: 18px;
+    }
+    .bold { font-weight: bold; }
+    tspan {
+      font: 14px SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace;
+      white-space: pre;
+      line-height: 18px;
+    }
+  </style>
+
+  <rect width="100%" height="100%" y="0" rx="4.5" class="bg" />
+
+  <text xml:space="preserve" class="container fg">
+    <tspan x="10px" y="28px"><tspan class="fg-bright-red bold">error[E0027]</tspan><tspan class="bold">: pattern does not mention fields `lineno`, `content`</tspan>
+</tspan>
+    <tspan x="10px" y="46px"><tspan>   </tspan><tspan class="fg-bright-blue bold"> ╭▸ </tspan><tspan>src/display_list.rs:139:32</tspan>
+</tspan>
+    <tspan x="10px" y="64px"><tspan>    </tspan><tspan class="fg-bright-blue bold">│</tspan>
+</tspan>
+    <tspan x="10px" y="82px"><tspan class="fg-bright-blue bold">139</tspan><tspan> </tspan><tspan class="fg-bright-blue bold">│</tspan><tspan>                           if let DisplayLine::Source {</tspan>
+</tspan>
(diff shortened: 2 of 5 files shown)


estebank:
Note that eventually we'd have to add a mechanism to provide highlighting exclusions (so that something like rustc can avoid highlighting trailing comments within the multiline span).

tests/color/ann_multiline_highlight_source.unicode.term.svg
epage, line 1: Huh, we don't have a `CONTRIBUTING.md` for this but something we've found very helpful is to add tests in the commit before, with them passing, showing the bad behavior. This makes it so when we diff for code changes, we also diff for the effect of that code that change.

epage (commented):
I'll probably leave any further reviewing to others but at least noticed this in pre-review that might help with the main review.

momomuchu:
Done. Restructured into two commits: the test first with pre-fix goldens showing the missing styling, then the fix with the goldens regenerated. Tree content is unchanged from the previous head, only the history was split.

src/renderer/render.rs
```diff
@@ -876,6 +876,39 @@ fn render_source_line(
...
+    // handles it for single-line annotations. Even when those fast paths aren't taken, the
+    // per-annotation styling logic matches on `MultilineStart`/`MultilineEnd` first to draw the
+    // horizontal connector, which made the `highlight_source` arm structurally unreachable for
+    // multiline annotations (see issue #427: `highlight_source` was silently ignored on them).
+    for ann in &line_info.annotations {
+        if !ann.highlight_source {
+            continue;
+        }
+        let (start_char, end_char) = match ann.annotation_type {
+            LineAnnotationType::Singleline => continue,
```
Muscraft, line 891:
```suggestion
            LineAnnotationType::Singleline => (ann.start.char, ann.end.char),
```
Note: You will need to remove the existing code for `SingleLine`:
https://github.com/rust-lang/annotate-snippets-rs/blob/c203ec6c38ca719b733c0fac3ea3e1cb8cfe7098/src/renderer/render.rs#L1192-L1200

src/renderer/render.rs
```diff
@@ -876,6 +876,39 @@ fn render_source_line(
...
+            // The span starts partway through this line and runs to the end of it.
+            LineAnnotationType::MultilineStart(_) => {
+                let rest_of_line = line_info.line[ann.start.byte..].chars().count();
+                (ann.start.char, ann.start.char + rest_of_line)
```
Muscraft, line 895: `ann.start.byte` is the byte offset from the start of the `source`, not the current line
```suggestion
                (ann.start.char, line_info.line.chars().count())
```

src/renderer/render.rs
```diff
@@ -876,6 +876,39 @@ fn render_source_line(
...
+            LineAnnotationType::MultilineLine(_) => (0, line_info.line.chars().count()),
+        };
+        let underline = renderer.decor_style.underline(ann.is_primary());
+        buffer.set_style_range(
+            line_offset,
+            (code_offset + start_char).saturating_sub(left),
+            (code_offset + end_char).saturating_sub(left),
+            underline.style,
+            ann.is_primary(),
+        );
```
Muscraft, line 909: The current code does not account for tabs (`\t`), and should do something similar to [`style_substitution_highlights`](https://github.com/rust-lang/annotate-snippets-rs/blob/c203ec6c38ca719b733c0fac3ea3e1cb8cfe7098/src/renderer/render.rs#L1961-L1962):
```suggestion
        let extra_width_start = extra_width_from_tabs(line_info.line, start_char);
        let extra_width_end = extra_width_from_tabs(line_info.line, end_char);
        buffer.set_style_range(
            line_offset,
            (code_offset + start_char + extra_width_start).saturating_sub(left),
            (code_offset + end_char + extra_width_end).saturating_sub(left),
            underline.style,
            ann.is_primary(),
        );

```

<details>
<summary>Possible regression test</summary>

```rust
use annotate_snippets::{
    AnnotationKind, Group, Level, Patch, Renderer, Snippet, renderer::DecorStyle,
};
use snapbox::{assert_data_eq, file};

#[test]
fn case() {
    let source = "prefix\tstart\n\tmiddle\nend";

    let report = &[Group::with_level(Level::ERROR).element(
        Snippet::source(source)
            .annotation(AnnotationKind::Primary.span(7..24).highlight_source(true)),
    )];

    let expected_ascii = file!["<test-name>.ascii.term.svg": TermSvg];
    assert_data_eq!(renderer.render(report), expected_ascii);

    let expected_unicode = file!["<test-name>.unicode.term.svg": TermSvg];
    let renderer = renderer.decor_style(DecorStyle::Unicode);
    assert_data_eq!(renderer.render(report), expected_unicode);
}

File: src/renderer/render.rs [hunk 4]
```diff
@@ -876,6 +876,39 @@ fn render_source_line(
...
     }
 
+    // Apply `highlight_source` styling for multiline annotations here, before the fast paths
+    // below (the "short start" case just below, and the "only connector lines" case further
+    // down) get a chance to return without ever reaching the per-annotation styling logic that
+    // handles it for single-line annotations. Even when those fast paths aren't taken, the
+    // per-annotation styling logic matches on `MultilineStart`/`MultilineEnd` first to draw the
+    // horizontal connector, which made the `highlight_source` arm structurally unreachable for
+    // multiline annotations (see issue #427: `highlight_source` was silently ignored on them).
+    for ann in &line_info.annotations {
```
Muscraft, line 886: While playing around with these changes, I realized that when `.fold(false)` is set, lines that would normally be folded are not highlighted (when they should be). I suspect that [`annotated_lines`](https://github.com/rust-lang/annotate-snippets-rs/blob/c203ec6c38ca719b733c0fac3ea3e1cb8cfe7098/src/renderer/source_map.rs#L144) is the source of the issue, but I have not checked.

<details>
<summary>Possible regression test</summary>

```rust
use annotate_snippets::{
    AnnotationKind, Group, Level, Patch, Renderer, Snippet, renderer::DecorStyle,
};
use snapbox::{assert_data_eq, file};

#[test]
fn case() {
    let source = r#"line
line
line
line
normally folded
normally folded
normally folded
line
primary end
normally folded
normally folded
normally folded
line
context end
"#;

    let report = &[Group::with_level(Level::ERROR).element(
        Snippet::source(source)
            .fold(false)
            .annotation(AnnotationKind::Primary.span(1..84).highlight_source(true))
            .annotation(AnnotationKind::Context.span(0..149).highlight_source(true)),
    )];

    let expected_ascii = file!["<test-name>.ascii.term.svg": TermSvg];
    assert_data_eq!(renderer.render(report), expected_ascii);

    let expected_unicode = file!["<test-name>.unicode.term.svg": TermSvg];
    let renderer = renderer.decor_style(DecorStyle::Unicode);
    assert_data_eq!(renderer.render(report), expected_unicode);
}
```

</details>

Muscraft (changes requested):
Thanks for taking the time to fix this issue! I'm sorry for the extremely long delay in reviewing this! My life has been very busy recently, leaving me with very little time to review this properly.

PR #443, opened 2026-07-14
feat: Add no graphics renderer mode

### LINKED ISSUE #351: Screen reader mode
This is dependent on cargo/rustc wanting it

Users of applications that use miette can get a screen reader mode for diagnostics using `NO_GRAPHICS=`. ([source](https://hachyderm.io/@zkat@toot.cat/115584771253790503))

Image

While how it is configured would be up to
- `--error-format` flag in rustc
- `term.<something>` in cargo (which maybe could also read `NO_GRAPHICS`)

the idea of making output compatible with screen readers seems beneficial

estebank:
Introduce a new text only renderer mode. The new output is meant to be less verbose, and usable by users relying on Text-To-Speech and Braille displays. All information that annotate-snippets has available is surfaced, but with an eye at reducing duplicated output and making sure there is no reliance on visuals for highlighting or differentiating necessary information, beyond newlines.

Each individual error group is separated by a single newline. Within an error, each section follows the pattern "LEVEL: message" on one line, followed by the position in the format `at PATH:LL:CC`, indented by one space. If that position has a `Message`, its text is on that same line, separated by a colon: `at PATH:LL:CC: message`. If there are multiple `Message`s, every subsequent position from the same path skip printing it, we use a more textual format and is indented by one additional space: `on line LL, column CC: message`. For suggestions, if there are multiple, we render each of them on their own line.

The renderer is color aware, so the same stylesheet is used as regular ASCII and Unicode drawing modes.

example of multiple multiline removal rendering in no_graphics mode

Fixes #351.

### DIFF
--- a/src/renderer/no_graphics.rs
+++ b/src/renderer/no_graphics.rs
@@ -1,0 +1,394 @@
+use alloc::{string::String, vec::Vec};
+use core::cmp::Reverse;
+use core::fmt::{self, Write};
+
+use crate::{
+    Id, Renderer, Report,
+    renderer::{
+        ElementStyle, Stylesheet,
+        render::{MessageOrTitle, PreProcessedElement, PreProcessedGroup, pre_process},
+        source_map,
+    },
+};
+
+/// Print out a file position optimized for the data available.
+///
+/// When the `path` is available, we print `at PATH:LL:CC`, which is detected by terminals and
+/// editors as a path to a specific place in a file. Otherwise, we print`on line LL, column CC`,
+/// for the elements that are available.
+fn render_path(
+    output: &mut String,
+    path: Option<&str>,
+    line: Option<usize>,
+    col: Option<usize>,
+) -> Result<(), fmt::Error> {
+    match (path, line, col) {
+        (Some(path), Some(line), Some(col)) => {
+            // `at $DIR/file.txt:LL:CC`
+            write!(output, "at {path}:{line}:{col}")?;
+        }
+        (Some(path), Some(line), None) => {
+            // `at $DIR/file.txt:LL`
+            write!(output, "at {path}:{line}")?;
+        }
+        (Some(path), None, None) => {
+            // `at $DIR/file.txt`
+            write!(output, "at {path}")?;
+        }
+        (None, Some(line), Some(col)) => {
+            // We are not printing the path, so instead we show
+            // ` on line LL, column CC`.
+            write!(output, "on line {line}, column {col}")?;
+        }
+        (None, Some(line), None) => {
+            // ` on line LL`
+            write!(output, "on line {line}")?;
+        }
+        (None, None, Some(col)) => {
+            // ` on column CC`
+            write!(output, "on column {col}")?;
+        }
+        (Some(path), None, Some(line)) => {
+            // This condition should never have been called.
+            write!(output, "at {path}, line {line}")?;
+        }
+        (None, None, None) => {}
+    }
+    Ok(())
+}
+
+/// Render all errors from a `Report` as text only.
+///
+/// We will render output as follows:
+///
+/// ```text
+/// error EXXXX: main message
+///  at $DIR/file.ext:LL:CC: span label
+///   on line LL, column KK: span label
+/// note: note message
+///  at $DIR/file.ext:LL:CC
+/// help: suggestion message
+///   on line LL, add `suggestion`
+/// ```
+pub(crate) fn render_no_graphics(
+    renderer: &Renderer,
+    groups: Report<'_>,
+) -> Result<String, fmt::Error> {
+    let mut output = String::new();
+    let (_max_line_num, og_primary_path, groups) = pre_process(groups);
+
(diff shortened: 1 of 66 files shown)


estebank:
Added a commit to use the "regular" path:line:col format, purely so that we have it around for playing around and sharing the output with people.

epage:
Note that we ask for commits to be atomic, including representing how they should be reviewed and merged, rather than how they were developed. For more on commit structure, see https://epage.github.io/dev/pr-style/

src/renderer/no_graphics.rs
```diff
@@ -0,0 +1,275 @@
...
+            match section {
+                PreProcessedElement::Message(message) => {
+                    if last_suggestion_path.is_some() && message.level.name == Some(None) {
+                        let text = message.text.as_ref();
+                        let text = text.strip_prefix("and ").unwrap_or(text);
+
+                        buffer.append(line, "  or ", ElementStyle::NoStyle);
+                        buffer.append(line, text, ElementStyle::NoStyle);
+                        last_suggestion_path = None;
+                    } else {
```
epage: This branch feels very specialized in what we are pattern matching against

src/renderer/render.rs
```diff
@@ -2620,7 +2620,7 @@ enum PreProcessedElement<'a> {
     Padding(Padding),
 }
 
-fn pre_process<'a>(
+pub(crate) fn pre_process<'a>(
```
epage, line 2624: Should we pull pre-processing out?

src/renderer/no_graphics.rs
```diff
@@ -0,0 +1,275 @@
...
+    let mut iter = groups.into_iter().peekable();
+    while let Some(PreProcessedGroup {
+        group,
+        elements,
+        primary_path: _,
+        max_depth: _,
+    }) = iter.next()
+    {
+        let mut buffer = StyledBuffer::new();
+        let mut line = 0;
```
epage: Whats `line`?

epage: Oh, the current line in the buffer?

Do we need `StyledBuffer` for this mode?

src/renderer/no_graphics.rs
```diff
@@ -0,0 +1,275 @@
...
+            id: Some(id),
+            url: _,
+        }) = &title.id()
+        {
+            // error EXXXX: message
+            //       ^^^^^
+            buffer.append(*line, " ", ElementStyle::NoStyle);
+            buffer.append(*line, id, ElementStyle::NoStyle);
+            label_width += 1 + id.len();
+        }
```
epage, line 376: We're rendering color but not hyperlinks?

estebank: We'll have to verify what TTS behavior is in all major platforms with OSC-8 links enabled. Same for paths. Scott and I are leaning towards having "path.rs, line LL, column, CC" instead of "path.rs:LL::CC" because it will always read correctly while the latter can read on some systems as a time, but the latter tends to work better with auto-linking machinery, like opening an editor to that place. We'll need feedback from accessibility experts to decide on these.

estebank:
Please be aware that I pushed this as a draft PR with the expectation that the code is not up-to-par. It is a bunch of my changes on top of Scott's work that we are still iterating on, but I feel it's at a stage where we can have wider, structured feedback and opening a PR felt like a natural place for that discussion. I plan to rewrite the commit history and make this non-draft once all three of us are satisfied with the general shape of things.

I see there were no comments on the output itself, I hope that's a good sign that the route we took is non-controversial?

src/renderer/no_graphics.rs
estebank: I think that emitting colors is reasonable: TTS should not be affected (other than calling out bold, IIRC?), but there are users that have poor eye-sight but still *can* see some, and the color would help for that kind of "mixed media" consumption.

I'm hearing from people who really want this output too for reduced verbosity reasons, so it feels like we can have a two-for-one here. Removing colors is still an option for the end user, but adding colors that aren't there isn't :)

src/renderer/no_graphics.rs
estebank: I'd tend to agree.

src/renderer/no_graphics.rs
epage: We can have color without `StyledBuffer`. Really, the big benefit to `StyledBuffer` is that it is more of an editable buffer representing the screen but that comes at a cost (and yes, users like `ruff` are impacted by the rendering performance of annotate-snippets). The no-graphics mode is much easier to build up incrementally without such screen buffer.

estebank:
For reference, [I forced this patch on the rustc codebase](https://github.com/rust-lang/rust/commit/be4fac5987f40ee5d46bddd6522a72960423383f), and the output looks better than I could have hoped.

long-type-E0308 short-color

It makes me reconsider a little bit whether repeating the path on every span label would be reasonable or not so that they remain clickable (at the cost of more text for TTS to munch through).

PR #462, opened 2026-08-06
feat(render): Add Snippet::line_numbering(false) 

### LINKED ISSUE #383: Snippets without line numbers
Some times you want to use a snippet as a way to quote content, where
- line numbers aren't relevant
- it doesn't participate in the "primary" calculations

e.g. https://github.com/rust-lang/cargo/pull/16711

epage:
Having a `true` default happens with `fold` so I figured it would be
fine here.

`line_numbering` was inspired by [zola](https://www.getzola.org/documentation/content/syntax-highlighting/#annotations):

> `linenos` to enable line numbering.

Fixes #383

### DIFF
--- a/tests/ruff.rs
+++ b/tests/ruff.rs
@@ -1,4 +1,4 @@
-use annotate_snippets::{AnnotationKind, Group, Level, Renderer, Snippet};
+use annotate_snippets::{Annotation, AnnotationKind, Group, Level, Renderer, Snippet};
 
 use annotate_snippets::renderer::DecorStyle;
 use snapbox::{assert_data_eq, str};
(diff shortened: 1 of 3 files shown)


tests/ruff.rs
```diff
@@ -53,11 +54,11 @@ info: incompatible return types: `NotBoolable` is not assignable to `bool`
...
-1 | def __eq__(self, other: object) -> bool:
-2 |         if not isinstance(other, A):
-3 |         if not isinstance(other, A):
-4 |             return False
-5 |         return <logic to compare two `A` instances>
+  | def __eq__(self, other: object) -> bool:
+  |         if not isinstance(other, A):
+  |         if not isinstance(other, A):
+  |             return False
+  |         return <logic to compare two `A` instances>
```
Muscraft, line 61: Shouldn't the indentation be removed here?

Muscraft (approved):

tests/ruff.rs
epage: indentation is calculated for the whole report so the loss of indentation is not covered by these tests. The code guarding the max_line_num calculations is in there though.

Muscraft (approved):

tests/ruff.rs
```diff
@@ -0,0 +1,88 @@
...
+        Level::HELP
+            .secondary_title(
+                "It is recommended for `__eq__` to work with arbitrary objects, for example:",
+            )
+            .element(
+                Snippet::<Annotation<'_>>::source(
+                    "\
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, A):
+        if not isinstance(other, A):
```
estebank, line 37: This duplication doesn't seem intentional :)

epage: Fixed in https://github.com/rust-lang/annotate-snippets-rs/pull/484

